PR #57, opened 2018-01-13
Add middleware support

josephmancuso:
@afdolriski @ondoheer  What do you think of this implementation of middleware?

So each middleware can have a simple `before` and `after` method which will get called depending on if it needs to be called before or after a route. 

In the config file there are two different types of middleware settings. A single list which will run on every single request (if a route match was found) and a route middleware which will only register names into the dictionary which can then be used in individual routes like:

```python
Get().route('/home', 'DashboardController@show').middleware('auth')
```

Look inside the AuthMiddleware class in the file changes to see how middleware is made. This will also have a `craft middleware AuthenticationMiddleware` craft command


josephmancuso:
Should rename the middleware. Names were arbitrary names I chose. The authentication middleware doesn't really authenticate and the route middleware does some authentication checking

josephmancuso:
documentation https://github.com/josephmancuso/masonite/blob/52ad2bc0c6cc3c6a28b8ab92fd320db44bdae8bc/docs/Middleware.md

josephmancuso:
### Changed

Middleware is now a string. I think this will speed development as you won't need to import all the middleware you need.

josephmancuso:
## Changed

Renamed middleware to be more appropriate for what it does

josephmancuso:
Looks good to me. Will accept another PR if someone thinks changes are needed. I don't really grasp the concept of other frameworks implementations of middleware and "onions". I understand the concept but I'm not sure why its needed. Someone please enlighten me if our middleware needs to return responses like other middleware implementations

PR #75, opened 2018-02-01
Service Container Support

josephmancuso:
Add support for an IOC service container and auto resolving functions


josephmancuso:
This will bump the new version to 1.1

PR #102, opened 2018-08-14
Windows Installation

vegax87:


vegax87:
They can use Python 3.4+ without any problem, I just tested the compatibility with Python 3.7

josephmancuso:
ahh ok. I'll merge this in later today :)

PR #117, opened 2018-09-24
Add Middleware for Email Verification feature

mitchdennett:
Updated User migration and added middleware.

Relies on XXX PR for core

### DIFF
--- a/app/http/middleware/VerifyEmailMiddleware.py
+++ b/app/http/middleware/VerifyEmailMiddleware.py
@@ -1,17 +1,29 @@
-''' Verify Email Middleware '''
+""" Verify Email Middleware """
 from masonite.request import Request
 
 class VerifyEmailMiddleware:
-    ''' Verify Email Middleware '''
+    """Middleware To Check If The User Has Verified Their Email
+    """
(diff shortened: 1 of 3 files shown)


config/middleware.py
```diff
@@ -40,4 +41,5 @@
 
 ROUTE_MIDDLEWARE = {
     'auth':  AuthenticationMiddleware,
+    'verify_email': VerifyEmailMiddleware
```
josephmancuso: let's rename this to just `verified`

app/http/middleware/VerifyEmailMiddleware.py
```diff
@@ -0,0 +1,17 @@
+''' Verify Email Middleware '''
+from masonite.request import Request
+
+class VerifyEmailMiddleware:
+    ''' Verify Email Middleware '''
+
+    def __init__(self, request: Request):
+        self.request = request
```
josephmancuso: Let's just clean this class up with the new docstrings we are using with other classes that look something like:

```
"""Short description here

Arguments:
    {something} here

...
```

PR #126, opened 2018-11-06
cleanup unused imports and whitespace

resmo:
cleaned up a bit but really no bad feelings if we keep it as is.


josephmancuso:
@resmo awesome! Thanks for contributing!

PR #154, opened 2019-09-19
Add shebang to craft script

carc1n0gen:
This allows direct execution of the craft script via `./craft` instead needing to do `python craft`. The latter will still work though.


josephmancuso:
Thanks! :)

PR #168, opened 2019-11-03
Feature/170 - Add contracts and interfaces

### LINKED ISSUE #170: Add Masonite Interfaces to Masonite Contract classes
Need to use the Masonite interfaces package for Masonite 2.3 contract classes

josephmancuso:
Closes #170


josephmancuso:
petty sure this isn't possible. Can't get it to work when resolving .. For some reason the inspect module doesn't correctly detect the signature of an object that has an interface .. will close

PR #180, opened 2019-11-04
Add pre-commit hooks

josephmancuso:
- Add additional tools require for pre-commit.
- Fix existing issues.


josephmancuso:
The more i think of this, the more I am not sure these things can be added directly to the repo 🤔 . These might just have to be linters and hooks within the actual CI pipeline since this adds a whole lot of meta files

mandarvaze:
The point of adding these files to the repo is because this repo forms the basis of the new projects, right ?

So my idea was that by adding the linters and hooks, we can ensure that new projects created are of high quality.

Let me know if you wish me to continue working on this.
If not, you can close this PR. (without merging)

josephmancuso:
Can all that be done without adding a lot of meta files?

Just seems to me like we would be cluttering new projects with these meta files. I really like the idea of generating high quality looking applications but can we do it without all the `.*` files?

We are also going to be adding more files like a `package.json` file for npm and stuff so i just don't want new projects getting too crowded

mandarvaze:
Lot of these tools look for their config files at either `pwd`, `$HOME` or `/etc` (Last two definitely not an option for us)

I can try and see if we can move all the dot files into a separate folder say `config`, instead of root of the project, If I am unable to do so - I will close this PR myself. Make sense ?

mandarvaze:
BTW, having meta files (as you call them) in root of the project is not uncommon.
See https://github.com/sdispater/poetry, https://github.com/Pylons/pyramid - but I'll defer to your judgement.

mandarvaze:
@josephmancuso Should I add other linters like black/isort to github action ?
I saw that you are already doing some [linting](https://github.com/MasoniteFramework/masonite/pull/169/commits/dda8a107ac147536a3111aecbdd812d2f322d829)  on 3.7, so `black` might work.

josephmancuso:
@mandarvaze whats the benefit to using black over other linters?

mandarvaze:
> @mandarvaze whats the benefit to using black over other linters?

black provides consistent code formatting (You've used it, so you know)
Using it as a pre-commit hook ensures that contributors do not need to "remember" to format using black. When they attempt the commit, black will actually **fix** it for them, but hook "fails" so no commit is made. The contributors then just need to `git add` the modified-by-black files to the commit)

Same goes for `isort`

`flake8` OTOH, will just report errors like missing doc string for example - which it can't "fix for you"

This PR adds all three. (and few more like `codespell`)

josephmancuso:
Ok i see. I am coming around to having the meta files in there. Why are all the tests failing though?

mandarvaze:
> Why are all the tests failing though?

1. "lint" needs to run only in 3.6 onwards. I can fix that. Orator does have similar step.
2. Cause `.pre-commit-config.yml` refers to the files inside `.venv` - which does not exists on travis environment resulting in errors like : `Executable `.venv/bin/flake8` not found`

mandarvaze:
I'm closing this, since it was created before `core` was merged. Now this branch is badly out of date. I'm creating another branch with these changes, plus more flake8 error fixes (after bringing in files from `core`) and then will submit another PR.

PR #197, opened 2019-11-25
Add support for Presets

### LINKED ISSUE #183: Add built-in support for Vue
Hi 
Thanks for your greate work.
It would be greate if Masonite integrate with vue js out of the box.Just like Laravel supports the vue.

lazyguru:
Fixes #183 

I'm not done here yet as I still need to add in some more unit tests ~and the methods could all use comments too~.  This is more or less a direct port from Laravel and as I am not as familiar with Python as I am with PHP would appreciate ~early~ feedback on it.

NOTE: I targeted develop here but in hindsight, I don't think there are any breaking changes so I could just as easily rebase and target 2.2 instead if you want.


package.json
```diff
@@ -0,0 +1,21 @@
...
+        "watch": "npm run development -- --watch",
+        "watch-poll": "npm run watch -- --watch-poll",
+        "hot": "cross-env NODE_ENV=development node_modules/webpack-dev-server/bin/webpack-dev-server.js --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",
+        "prod": "npm run production",
+        "production": "cross-env NODE_ENV=production node_modules/webpack/bin/webpack.js --no-progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js"
+    },
+    "devDependencies": {
+        "axios": "^0.19",
+        "cross-env": "^5.1",
+        "laravel-mix": "^4.0.7",
```
mandarvaze: Is this needed ?

src/masonite/commands/PresetCommand.py
```diff
@@ -0,0 +1,45 @@
...
+    def remove(self):
+        """Removes frontend scaffolding"""
+        Remove().install()
+        self.info('Frontend scaffolding removed successfully.')
+
+    def react(self):
+        """Add React frontend while also removing Vue (if it was previously selected)"""
+        React().install()
+        self.info('React scaffolding installed successfully.')
+        self.comment('Please run "npm install && npm run dev" to compile your fresh scaffolding.')
```
mandarvaze: Is it always `dev` ?

src/masonite/commands/presets/Remove.py
```diff
@@ -0,0 +1,51 @@
...
+import shutil
+from ..presets import Preset
+
+
+class Remove(Preset):
+    """Removes any defined Preset"""
+
+    def install(self):
+        """Install the preset"""
+        self.update_packages()
```
mandarvaze: Are these three needed here ? Especially since we are removing stuff.

josephmancuso:
Let's keep it based on develop. Theres no breaking changes but eh, let's make this a 2.3 feature. 

Usually when we get near the release of a next major I just throw a bunch of things in develop

package.json
lazyguru: To use the webpack.mix.js stuff, yes. Laravel Mix is a webpack plug-in that isn’t really specific to Laravel, it was just created for Laravel by someone in the Laravel community

src/masonite/commands/PresetCommand.py
lazyguru: Yeah, I thought that was weird too. We could probably remove it and just hard-code it to dev always.

src/masonite/commands/presets/Remove.py
lazyguru: I’m not sure I understand? Those functions all do something. So they should be there and they are not no-ops

src/masonite/commands/PresetCommand.py
lazyguru: It would be `dev` or `watch` in development, but specifying `dev` here makes the most sense

PR #216, opened 2019-12-08
Add validation to authentication quickstart - Fix/173

girardinsamuel:


src/masonite/commands/RuleEnclosureCommand.py
```diff
@@ -0,0 +1,17 @@
...
+class RuleEnclosureCommand(BaseScaffoldCommand):
+    """
+    Creates a new rule enclosure.
+
+    rule:enclosure
+        {name : Name of the rule enclosure}
+    """
+
+    scaffold_name = 'Rule'
+    postfix = ""
```
RoySegall: Should be:
```suggestion
    postfix = ''
```
Since all the other vairables are assinged with a single and not a dobule quote.

src/masonite/snippets/auth/templates/auth/login.html
```diff
@@ -32,4 +46,4 @@ <h5 class="mb-0">Login</h5>
         </div>
     </div>
 </div>
-{% endblock %}
+{% endblock %}
```
RoySegall: An empty line in the missing of the file is missing.

girardinsamuel: Done !

josephmancuso:
made a few changes and merged 👍 looks good!

PR #217, opened 2019-12-08
Rename auth controllers and templates - Fix/182

girardinsamuel:


josephmancuso:
Too many conflicts now in this branch. Let's revisit this soon. It won't be able to be finished until a next first of Masonite anyway and the next major releases tomorrow

PR #218, opened 2019-12-09
Add Tailwind CSS preset

lazyguru:
This is a start for adding Tailwind CSS support by porting most of https://github.com/laravel-frontend-presets/tailwindcss


lazyguru:
@josephmancuso there is more needed here.  This PR just added the setup, but the templates should be replaced with ones that use Tailwind instead of Bootstrap

josephmancuso:
oh woops. Reverted. Just open a new PR then

PR #232, opened 2020-01-26
added quote cleaning

### LINKED ISSUE #231: Improve cleaning of request input
**Is your feature request related to a problem? Please describe.**
The problem is that when submitting description style content that could be in the form of HTML, the cleaning is too strict. We are cleaning things like quotes which could not be cleaned.

**What do we currently have to do now?** 
we have the cleaning for the request input like this:

```python
request.input('something', clean=False)
```

**Describe the solution you'd like**

we should have something like this:

```python
request.input('something', quote=False)
```

This will decide on whether it should clean quotes

- [ ] Is this a breaking change?
no

josephmancuso:
can now decide on if you want to clean quotes when cleaning input

Closes #231

### DIFF
--- a/src/masonite/helpers/misc.py
+++ b/src/masonite/helpers/misc.py
@@ -39,7 +39,7 @@ def dot(data, compile_to=None):
     return compiling
 
 
-def clean_request_input(value, clean=True):
+def clean_request_input(value, clean=True, quote=True):
     if not clean:
         return value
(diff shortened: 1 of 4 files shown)


mandarvaze:
May be add test cases for `quote=True` ? I know it is default but it would ensure both the paths are covered.

PR #233, opened 2020-01-29
Add redirect to intended route functionality

carc1n0gen:
Previously in the auth scaffold upon successful login there was a hard coded redirect

```
if auth.login(request.input('email'), request.input('password')):
    return request.redirect('/home')
```

and with this change, the same can be done like this:

```
if auth.login(request.input('email'), request.input('password')):
    return request.intended(default='/home')
```

This will look for a session value `__intended` and if it exists, redirect there, otherwise fallback to the whatever is passed as default.

I will also make a PR to the cookie cutter repo to update the auth middleware to set `__intended` right before redirecting to login . EDIT: said PR https://github.com/MasoniteFramework/cookie-cutter/pull/5


src/masonite/request.py
```diff
@@ -785,25 +785,33 @@ def reset_redirections(self):
...
         """Return a URI for redirection depending on several use cases.
 
         Keyword Arguments:
             default {string} -- Default value if nothing can be found. (default: {None})
+            intended {boolean} -- Whether or not to redirect to a previously accessed route that required auth. (default: {False})
 
         Returns:
             self
         """
         self.with_input()
```
carc1n0gen: Maybe I should move this to the else block of the `if intended` check so that the intended redirect doesn't get side effects of with_input

mandarvaze:
@carc1n0gen I looked at the tests - What is the difference between `request.with_intended` and `request.intended(default='..')`

May be add documentation changes to this PR - explaining when one might choose one style over other ?

carc1n0gen:
@mandarvaze .with_intended sets the __intended session value.  The default kwarg to .intended is a fallback to redirect to in case the __intended session value does not exist.  Yes I agree this needs documentation

josephmancuso:
@carc1n0gen This is ready to review and test right?

josephmancuso:
oh you requested review. I see. Disregard my dumb question

josephmancuso:
Made some changes to this PR. 

The functionality working right now is you will redirect intended like so:

```python
HomeController:

    show:
        if not request.user():
           return request.redirect('/login').intend() # this sets up the intended redirect
```

Then in the login controller (after they login successfully):

```python
LoginController:

    store:
        if not request.user():
           return request.redirect_intended(default='/home') # this either redirects to the intended path or redirects to home.
```

josephmancuso:
I modified it slightly to be this now:

```python
HomeController:

    show:
        if not request.user():
           return request.redirect('/login').then_back() # this sets up the intended redirect
```

Then in the login controller (after they login successfully):

```python
LoginController:

    store:
        if not request.user():
           return request.redirect_intended(default='/home') # this either redirects to the intended path or redirects to home.
```

josephmancuso:
I didn't like the `request.redirect('/dashboard').intend()`. It wasn't very clear. I think changing it to `request.redirect('/dashboard').then_back()`. is much clearer

PR #247, opened 2020-02-14
Fix/246

holic-cl:
Hi, this is a fix for prevent the double encode a stored password (which is happening with PostgreSQL).


holic-cl:
Hi, i found the problem, i was saving the user password in this way:
```
password = bcrypt.hashpw(
    password.encode(encoding='UTF-8', errors='strict'),
    bcrypt.gensalt()
)
data = {'name': "user", 'email': "email@example.com", 'password': password}
User.create(data)
```

To fix this issue i had to decode the password before save the new user
```
data = {'name': "user", 'email': "email@example.com", 'password': password.decode('utf-8')}
User.create(data)
```

PR #255, opened 2020-03-04
fixed issue for ably broadcast driver

abhijitmamarde:
fixed issue for ably broadcast driver: 'BroadcastConfig key was not found in the container'

exception log:

```
Traceback (most recent call last):
  File "/home/abhi/pwork/masonite_projs/todo/.py38env/lib/python3.8/site-packages/masonite/managers/Manager.py", line 77, in create_driver
    self.manage_driver = self.container.make(
  File "/home/abhi/pwork/masonite_projs/todo/.py38env/lib/python3.8/site-packages/masonite/app.py", line 89, in make
    obj = self.resolve(obj, *arguments)
  File "/home/abhi/pwork/masonite_projs/todo/.py38env/lib/python3.8/site-packages/masonite/app.py", line 190, in resolve
    return obj(*objects)
  File "/home/abhi/pwork/masonite_projs/todo/.py38env/lib/python3.8/site-packages/masonite/drivers/broadcast/BroadcastAblyDriver.py", line 18, in __init__
    self.config = app.make('BroadcastConfig')
  File "/home/abhi/pwork/masonite_projs/todo/.py38env/lib/python3.8/site-packages/masonite/app.py", line 100, in make
    raise MissingContainerBindingNotFound(
masonite.exceptions.MissingContainerBindingNotFound: BroadcastConfig key was not found in the container
```

This PR is now raised on correct repo - masonite and correct branch, earlier raised incorrectly on core repo (old PR#947).

Let me know if any more info required to accept this PR.

Thanks!

### DIFF
--- a/src/masonite/drivers/broadcast/BroadcastAblyDriver.py
+++ b/src/masonite/drivers/broadcast/BroadcastAblyDriver.py
@@ -10,7 +10,7 @@ from ...helpers import config
 class BroadcastAblyDriver(BroadcastContract, BaseDriver):
     """Class for the Ably Driver."""
 
-    def __init__(self, app: App):
+    def __init__(self):
         """Ably driver constructor.
 
         Arguments:
@@ -53,6 +53,9 @@ class BroadcastAblyDriver(BroadcastContract, BaseDriver):
                 'Could not find the "ably" library. Please pip install this library running "pip install ably"')
 
         configuration = config('broadcast.drivers.ably')
+        if not configuration:
+            raise Exception('Could not find ably broadcast configuration')
+
         client = AblyRest('{0}'.format(
             configuration['secret']
         ))
--- a/src/masonite/drivers/broadcast/BroadcastAblyDriver.py
+++ b/src/masonite/drivers/broadcast/BroadcastAblyDriver.py
@@ -3,7 +3,6 @@
 from ...contracts import BroadcastContract
 from ...drivers import BaseDriver
 from ...exceptions import DriverLibraryNotFound
-from ...app import App
 from ...helpers import config
 
 
--- a/src/masonite/drivers/broadcast/BroadcastAblyDriver.py
+++ b/src/masonite/drivers/broadcast/BroadcastAblyDriver.py
@@ -4,6 +4,7 @@ from ...contracts import BroadcastContract
 from ...drivers import BaseDriver
 from ...exceptions import DriverLibraryNotFound
 from ...app import App
+from ...helpers import config
 
 
 class BroadcastAblyDriver(BroadcastContract, BaseDriver):
@@ -15,7 +16,6 @@ class BroadcastAblyDriver(BroadcastContract, BaseDriver):
         Arguments:
             BroadcastConfig {config.broadcast} -- Broadcast configuration setting.
         """
-        self.config = app.make('BroadcastConfig')
         self.ssl_message = True
 
     def ssl(self, boolean):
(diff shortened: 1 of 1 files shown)


mandarvaze:
@abhijitmamarde Since `app` is not used anymore, may be remove it (Error reported by deepsource)

Is there a test case that could have caught this ? 
(If easy) May be add a test so that it does not break in the future.

src/masonite/drivers/broadcast/BroadcastAblyDriver.py
```diff
@@ -52,8 +52,9 @@ def channel(self, channels, message, event='base-event'):
             raise DriverLibraryNotFound(
                 'Could not find the "ably" library. Please pip install this library running "pip install ably"')
 
+        configuration = config('broadcast.drivers.ably')
         client = AblyRest('{0}'.format(
-            self.config.DRIVERS['ably']['secret']
+            configuration['secret']
```
josephmancuso: This won't work. If the ably config doesn't exist then configuration will return `None`

vaibhavmule: @josephmancuso 

Will it work?

```suggestion
            configuration['secret'] or []
```

josephmancuso: No that won't work. configuration would be None and you can't get `['secret']` from None

josephmancuso: I beleive `config` takes a second parameter which is a default argument. So maybe `config('dot.notation.here', '')` would work

josephmancuso: It should probably throw an error though instead since you can't even use it without a secret set

mandarvaze: BroadcastPushDriver has similar code [here](https://github.com/MasoniteFramework/masonite/blob/2.3/src/masonite/drivers/broadcast/BroadcastPusherDriver.py#L54)

Either both work (cause configuration is always available) or both need to be fixed IMO

abhijitmamarde: So whats the conclusion, can I just remove the unused app parameter in init, and raise the PR again?

If anything else required let me know!

mandarvaze: Let @josephmancuso comment.

josephmancuso: The conclusion is to do 

```python
        configuration = config('broadcast.drivers.ably')
        if not configuration:
            raise Exception('Could not find ably broadcast configuration')

        client = AblyRest('{0}'.format(
            self.config.DRIVERS['ably']['secret']
            configuration['secret']
```

abhijitmamarde: okay will check it once and will raise the PR, will update this same PR tomorrow :)

src/masonite/drivers/broadcast/BroadcastAblyDriver.py
```diff
@@ -53,6 +53,9 @@ def channel(self, channels, message, event='base-event'):
                 'Could not find the "ably" library. Please pip install this library running "pip install ably"')
 
         configuration = config('broadcast.drivers.ably')
+        if not configuration:
+            raise Exception('Could not find ably broadcast configuration')
+
```
abhijitmamarde: added the check as suggested by @josephmancuso

abhijitmamarde:
updated the code, added the suggested check by @josephmancuso

src/masonite/drivers/broadcast/BroadcastAblyDriver.py
abhijitmamarde: done. please reivew :)

josephmancuso:
This is good. I haven't had a lot of time to manually test this but I will get to it soon

PR #282, opened 2020-04-09
Add unit test coverage for shared-stubs

lazyguru:
This adds assertions for the bootstrap.js file that was added in #235 for vue and react presets from the shared-stubs directory.


lazyguru:
Weird? Why does 3.6 fail but the others all succeed?

lazyguru:
Added commit to #283

PR #283, opened 2020-04-10
Add missing app.scss for Vue/React presets

lazyguru:
This will add an empty app.scss to resources/sass when calling `craft preset vue` or `craft preset react`


josephmancuso:
LGTM 👍

Issue #87, opened 2018-06-16
Init install fails - [CommandNotFound] Command "key" is not defined. 


patvdleer:
Trying the initial test setup as displayed on the wiki I get the following;

PY3.6, `$ craft -V; Craft Version: 2.0.2`

```sh
patrick@patrick-P506:~/tmp$ craft new project
Crafting Application ...

Application Created Successfully!

Now just cd into your project and run

    $ craft install

to install the project dependencies.

Create Something Amazing!
patrick@patrick-P506:~/tmp$ cd project/
patrick@patrick-P506:~/tmp/project$ craft install

                                 
  [CommandNotFound]              
  Command "key" is not defined.  
                                 
```

josephmancuso:
what is the pip version?

josephmancuso:
I can't recreate it. Are you using pip 9 or pip 10? and Mac or Linuc?

josephmancuso:
Can you tell me your exact installation steps?

josephmancuso:
The latest version of craft shoulld be `2.0.3`.

try running:

```
$ pip3 install --upgrade masonite-cli
```

josephmancuso:
and also try running it under sudo of you are using UNIX

nmalacarne:
@josephmancuso is right. You need to run with sudo.

![image](https://user-images.githubusercontent.com/2746307/41503556-81317162-71a5-11e8-93fd-7c2802d63f2a.png)

josephmancuso:
@nmalacarne It's weird that it;'s throwing a key not defined error. What is the exact command you use? I keep trying to replicate this.

    $ pip3 install masonite-cli

?

josephmancuso:
Others coming here can read these docs to see if they help: https://docs.masoniteproject.com/prologue/known-installation-issues

nmalacarne:
I can reproduce it with the following commands:

```
sudo pip3 install masonite-cli \
&& craft new testing \
&& cd testing \
&& craft install
```
<details><summary>Output</summary>

```console
WARNING: Running pip install with root privileges is generally not a good idea. Try `pip3 install --user` instead.
Collecting masonite-cli
  Using cached https://files.pythonhosted.org/packages/7c/a8/47f745bde82b350d75d1f2d2a2f0a098ce369fd015c91b5e8729ee95f9df/masonite-cli-2.0.3.tar.gz
Collecting cleo==0.6.5 (from masonite-cli)
[...]

PR #295, opened 2020-05-25
fix/pip-installation: Fix installation of dependencies by using the Python module of pip

nok:
**Describe the bug**
By calling `craft new` all required packages will be installed automatically. In general it's a good idea, but this design decision has some drawbacks. In my case the requirements were installed successfully, but the wrong environment was chosen. So I got this consequential error:

```
[CommandNotFound]
Command "key" is not defined.
```

The sub command "key" could not be found, because the command "key" was not registered in this loop:

```python
# masonite/src/masonite/cli.py
try:
    from wsgi import container  # <----- ImportError in wsgi.py
    from cleo import Command
    for key, value in container.collect(Command).items():
        application.add(value)
except ErrorDuringImport as e:
    print(e)
except ImportError:  # <----- Catch and ignore the ImportError
    pass
```
This happen, because of some broken imports:

```python
# wsgi.py

from masonite.app import App

from bootstrap.start import app
from config import application, providers  # <----- ImportError in config/providers.py
```

```python
# config/providers.py
# ...

from masonite.logging.providers import LoggingProvider  # <----- Missing dependency "masionate-logging"
from masonite.validation.providers import ValidationProvider  # <----- Missing dependency masionate-validation
```

All this happen, because the dependencies were not installed to the right place. The fix is simple:

We just use the module `pip` of Python to install all requirements. The name of the executable isn't necessary. (There are many cases where `pip3` isn't or doesn't link to the right package manager.)

```shell
pip3 install -r requirements.txt           # old
python -m pip install -r requirements.txt  # new/fix
```

**To Reproduce**
1. `conda create -y -n maso_env python=3.7 && conda activate maso_env`
2. `pip install masonite`
3. `mkdir maso_app && cd maso_app`
4. `craft new`

**Expected behavior**
An installation without errors.

**Screenshots**
Screenshot 2020-05-25 at 15 02 15
(In the screenshot the name of the directory and conda environment are changed.)

**Desktop (please complete the following information):**
 - OS: macOS 10.15.4 
 - Version Python: 3.7.7
 - Version Masonite: 2.3.7

Best, Darius 😄

### DIFF
--- a/src/masonite/commands/InstallCommand.py
+++ b/src/masonite/commands/InstallCommand.py
@@ -34,9 +34,9 @@ class InstallCommand(Command):
             except Exception:
                 self.comment(
                     """Pipenv could not install from your Pipfile .. reverting to pip installing requirements.txt""")
-                call(["pip3", "install", "-r", "requirements.txt"])
+                call(["python", "-m", "pip", "install", "-r", "requirements.txt"])
         elif os.path.isfile('requirements.txt'):
-            call(["pip3", "install", "-r", "requirements.txt"])
+            call(["python", "-m", "pip", "install", "-r", "requirements.txt"])
         else:
             raise OSError(
                 'Could not find a Pipfile or a requirements.txt file')


nok:
In addition this change should also fix the past issues with `conda`:
https://github.com/MasoniteFramework/masonite/issues/87#issuecomment-398949833

PR #298, opened 2020-05-27
Feature/297

### LINKED ISSUE #297: Handing request body with an array as the root object
Currently `Request.request_variables` is expecting the input to be a `dict`
We should support lists as well.

```
curl -H 'Content-Type: application/json' \
    -d '[{"key": "val"}, {"item2": "val2"}]' \
    http://127.0.0.1:8000/test
```

How:

- [ ] Add a property `request.is_array` indicating whether the root object is an array or dict.
- [x] `request.all()` returns original list or dict depending on `request.is_array`
- [x] `request.input(name)` should support list indices as well.
- [x] extend `request.input(name)` so any of the following would be valid.

       `request.input(0)` first index in list.
       `request.input('0')` dict key `'0'` or list index `0`
       `request.input('0.foo.1')` == request_variables[0]['foo'][1]
       `request.input('0.foo.1')` == request_variables['0']['foo']['1']

josephmancuso:
* Fixes issue with null values not returning correctly when using dot notation
* Adds improvements for dot notation when working with lists (able to retrieve values with list indexes `key.0.value.1`)
* Fixes issue with getting a list as a payload throwing an exception
* `requst.all()` will Now returns a list if a list was passed into it

Closes #297

### DIFF
--- a/src/masonite/request.py
+++ b/src/masonite/request.py
@@ -84,7 +84,7 @@ class Request(Extendable):
             return self.request_variables.get(name)
 
         if '.' in name and isinstance(self.request_variables.get(name.split('.')[0]), dict):
-            return DictDot().dot(name, self.request_variables)
+            return clean_request_input(DictDot().dot(name, self.request_variables, default=default), clean=clean)
 
         elif '.' in name:
             name = dot(name, "{1}[{.}]")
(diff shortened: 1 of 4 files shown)


src/masonite/request.py
```diff
@@ -78,10 +79,12 @@ def input(self, name, default=False, clean=False):
         Returns:
             string
         """
+        name = str(name)
+        if name.isnumeric():
+            return self.request_variables.get(name)
```
ar45, line 84: Oops. Missed default here

PR #299, opened 2020-05-27
'Mail - mail_from and to_addresses

ar45:
- Make mail.send_from more consistent. Allow setting the mail from address including the display name.
	Supported using:
		`mail.send_from('john.smith@example.com')`
		`mail.send_from('John Smith <john.smith@example.com>')`
		`mail.send_from('john.smith@example.com', 'John Smith')`

- renamed `to_address` to `to_addresses` and is now held as a list of addresses
	Supported:
		`mail.to('john@example.com')`
		`mail.to(['john@example.com'])`
		`mail.to(['john@example.com', 'another@example.com'])`

### DIFF
--- a/src/masonite/drivers/mail/BaseMailDriver.py
+++ b/src/masonite/drivers/mail/BaseMailDriver.py
@@ -1,11 +1,15 @@
 """Base mail driver module."""
 
 import copy
+import re
 
 from ...app import App
 from ...helpers import config
 from ...response import Responsable
 from .. import BaseDriver
+
+
+MAIL_FROM_RE = re.compile(r'(?:"?([^"]*)"?\s)?(?:<?(.+@[^>]+)>?)')
 
 
 class BaseMailDriver(BaseDriver, Responsable):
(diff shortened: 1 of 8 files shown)


src/masonite/drivers/mail/BaseMailDriver.py
```diff
@@ -39,7 +52,10 @@ def to(self, user_email):
         if callable(user_email):
             user_email = user_email.email
 
-        self.to_address = user_email
+        if isinstance(user_email, (list, tuple)):
+            self.to_addresses = user_email
```
josephmancuso, line 56: Should this be `+=` maybe? it would make it able to be called several times

ar45: So how would you reset?

ar45:
I don't know how to read the DeepSource check and what it requires to change. If you'd like to push an update for that would be helpful. 
Thank you!

josephmancuso:
eh don't worry about deep source check. It usually gets out of hand and i'll spend a whole night every now and then fixing them all.

PR #300, opened 2020-06-01
Expose query string paraemters not only for GET requests.

ar45:
query string parameters are set as request_variables and subsequently overridden by post variables.

- `request.all()` and `request.input()` return combined query params and post params.
- Added method `request.query_param(name, default, multi_value=False)` to retrieve query string parameters.
- CONTENT_TYPE case insensitive

### DIFF
--- a/src/masonite/request.py
+++ b/src/masonite/request.py
@@ -11,6 +11,7 @@ of this class.
 import re
 from cgi import MiniFieldStorage
 from http import cookies
+from urllib.parse import parse_qs
 
 import tldextract
 from cryptography.fernet import InvalidToken
@@ -58,6 +59,7 @@ class Request(Extendable):
         self.request_variables = {}
         self._test_user = False
         self.raw_input = None
+        self.query_params = {}
 
         if environ:
             self.load_environ(environ)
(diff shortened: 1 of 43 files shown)


src/masonite/request.py
```diff
@@ -89,6 +91,29 @@ def input(self, name, default=False, clean=False, quote=True):
...
+        Keyword Arguments:
+            default {any} -- Default value if input does not exist (default: {None})
+            first {bool} -- Whether to return all values of a multi value query string param
+                            ie. quuery_param=one&query_param=two (default: {False})
+
+        Returns:
+            any
+        """
+        try:
+            value = self.query_params[name]
```
josephmancuso, line 109: This block can simply go to `value = self.query_params.get(name, default)`

ar45: If not multi_value, we return value [0] which is not what we want to do for default [0]. default is to be returned verbatim.

ar45:
I still have to add some tests for this.

josephmancuso:
I added a way to get all query params

ar45:
Thanks @josephmancuso

PR #302, opened 2020-06-03
Mail mimetype text

ar45:


src/masonite/drivers/mail/BaseMailDriver.py
```diff
@@ -40,6 +43,42 @@ def mail_from_header(self):
...
+    def message_body(self):
+        """Returns the body of the message
+        """
+        return self.html_content or self.text_content
+
+    @message_body.setter
+    @deprecated('Please use `.text_content` and `.html_content` instead.')
+    def message_body(self, value):
+        self.html_content = value
+
```
ar45, line 81: I wasn't sure whether to deprecate this or not. On the one hand, it is kind of useful to have a single method returning one of the available content.

src/masonite/drivers/mail/BaseMailDriver.py
```diff
@@ -40,6 +43,42 @@ def mail_from_header(self):
...
+    def html(self, content):
+        """Set the html content of the email.
+
+        Arguments:
+            content {string} -- The email html content.
+
+        Returns:
+            self
+        """
+        self.html_content = content
```
josephmancuso, line 67: self.text_content and self.html_content are repetitive. the point of setting one or the other is so later in the building process you can either set the html or plain text mime type.

src/masonite/drivers/mail/BaseMailDriver.py
```diff
@@ -40,6 +43,42 @@ def mail_from_header(self):
...
+
+    @property
+    @deprecated('Please use `.text_content` and `.html_content` instead.')
+    def message_body(self):
+        """Returns the body of the message
+        """
+        return self.html_content or self.text_content
+
+    @message_body.setter
+    @deprecated('Please use `.text_content` and `.html_content` instead.')
```
josephmancuso, line 78: I think we can do this without deprecating it. Just keep the setter property as message_body and we can detect the mimetype by a new property like `_mimetype`

src/masonite/drivers/mail/MailSmtpDriver.py
```diff
@@ -12,29 +12,38 @@
...
+        message['Reply-To'] = self.message_reply_to
+
+        if message_contents:
+            message_contents = MIMEText(message_contents, 'html')
+            message.attach(message_contents)
+
+        else:
+            if self.html_content:
+                message.attach(MIMEText(self.html_content, 'html'))
+
```
josephmancuso, line 29: The entire change here comes down to this if statement. We should keep it as `self.message_body` but do a check like:

```python
if self._mimetype == 'html':
    message.attach(MIMEText(self.message_body, 'html'))
elif self._mimetype == 'text':
    message.attach(MIMEText(self.message_body, 'plain'))

```

src/masonite/drivers/mail/BaseMailDriver.py
```diff
@@ -27,10 +29,11 @@ def __init__(self, app: App):
         self.to_addresses = []
         self.message_subject = 'Subject'
         self.message_reply_to = None
-        self.message_body = None
         self.from_name = self.config.FROM['name']
         self.from_address = self.config.FROM['address']
         self._queue = False
+        self.html_content = None
+        self.text_content = None
```
josephmancuso, line 36: We can do this and maintain the `message_body` attribute like we had. Simply manage a new `_mimetype` attribute instead

src/masonite/drivers/mail/BaseMailDriver.py
ar45: The 2 attributes are not the same thing at all.
The html content has rich html and markup and may even have different text, whereas the text content is a minimal text version including specific line quoting and chars which render differently.
Example 
```python
mail.html("<div>Hello&nbsp;<strong>User</strong></div>").text("Hello User\n message...").send()
```

src/masonite/drivers/mail/MailSmtpDriver.py
ar45: Again. I want to include both content types in the same message.

ar45: If anything I would deprecate the message_content arg. I only kept it to be backwards compatible.

src/masonite/drivers/mail/BaseMailDriver.py
josephmancuso: Oh ok. Right you want to use both. Gotcha

src/masonite/drivers/mail/MailSmtpDriver.py
josephmancuso: Right gotcha. You want to both both

src/masonite/drivers/mail/BaseMailDriver.py
ar45: I think we should deprecate setting `message_body` but allow the property to get whichever content type exists.

PR #310, opened 2020-06-27
fixed issue with session driver

josephmancuso:


tpow:
A minor suggestion: you could simplify line 106 of SessionCookieDriver from

```cookie_name = data[0].replace('s_', '').replace('f_', '').strip()```

to

```cookie_name = data[0].replace('f_', '').strip()```

since you are only working with the flashes in that block.

josephmancuso:
Nice catch. Does this work for you?

josephmancuso:
Nvm just saw your comment on the issue.

PR #314, opened 2020-07-05
Fix/311

### LINKED ISSUE #311: Can't use route compilers
I am trying to use the route compiler for year as described in the documentation. However, either there is something missing in the instructions, I'm missing something, or the `craft serve` command checks routes too early in the process (before the service providers are applied) and fails.

**To Reproduce**
I lifted the example from the documentation here: 
https://docs.masoniteproject.com/v/v2.3/the-basics/routing

Because I had previously added a service provider for a template view composer, I thought I understood what might be needed. However, the example is incomplete.

I ended up creating this in app/providers:
```python
from masonite.provider import ServiceProvider
from masonite.routes import Route


class RouteTypesProvider(ServiceProvider):
    """Provides a Route Compiler for various types"""

    wsgi = False

    def register(self):
        """Register objects into the Service Container"""
        pass

    def boot(self, route: Route):
        route.compile('year', r'([0-9]{4})')
```

I then edited config/providers.py, added the import 
`from app.providers.RouteTypesProvider import RouteTypesProvider`
and added RouteTypesProvider to the end of the PROVIDERS list.

Finally, I added a route that uses the year compiler. In this case I want the parameter also called year, so it looks like this:

`Get('/acct/?year:year', 'AccountController@show')`

I used `craft serve` to test it.

**Expected behavior**
I expected the server to start and be able to use the year compiler.

**Bug**
Despite having used `craft serve` successfully before, it now fails with an error AND worse, acts like the serve command doesn't exist:

```
problem in routes.web - InvalidRouteCompileException: Route compiler "year" is not an available route compiler. Verify you spelled it correctly or that you have added it using the compile() method.

                                   
  [CommandNotFound]                
  Command "serve" is not defined.  
```

**Note**
If I remove the :year from the route, it works again. Do route compilers not work? Did I miss a step? If I did, it would be nice to add this to the documentation. As it was, I had to make a few guesses. (The documentation for View Composers is helpfully more thorough https://docs.masoniteproject.com/v/v2.3/useful-features/view-composers-and-sharing )

josephmancuso:
Fixes #311

### DIFF
--- a/src/masonite/providers/RouteProvider.py
+++ b/src/masonite/providers/RouteProvider.py
@@ -31,7 +31,7 @@ class RouteProvider(ServiceProvider):
                 is found, execute that route and break out of the loop. We only need
                 one match. Routes are executed on a first come, first serve basis
             """
-            if matchurl.match(router.url) and request.get_request_method() in route.method_type:
+            if matchurl and matchurl.match(router.url) and request.get_request_method() in route.method_type:
                 route.load_request(request)
 
                 """Check if subdomains are active and if the route matches on the subdomain
--- a/src/masonite/routes.py
+++ b/src/masonite/routes.py
@@ -367,7 +367,7 @@ class BaseHttpRoute:
                         regex += Route.route_compilers[regex_route.split(':')[
                             1]]
                     except KeyError:
-                        if hasattr(self, '_compiled_regex'):
+                        if self.request:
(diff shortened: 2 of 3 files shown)


tpow:
It might be nice to include the tests for invalid routes (they appear to be commented out in the test) but I think that is mainly a test that the regex works so it may not be necessary.

PR #315, opened 2020-07-06
Add multiple arg support to dd()

tpow:
This adds support for passing multiple objects to dd() and having it show all the results. The existing single argument form still works.

I kept the members property since it was in the original exception_handler, but unless I'm missing something it isn't used.

I made some edits to the output. It includes a header showing the number of objects if you dump more than one. I also added separators between them so it is easier to tell where the output for one ends and the next one begins.

### DIFF
--- a/src/masonite/exception_handler.py
+++ b/src/masonite/exception_handler.py
@@ -109,8 +109,13 @@ class DD:
     def __init__(self, container):
         self.app = container
 
-    def dump(self, obj):
-        self.app.bind('ObjDump', obj)
+    def dump(self, *args):
+        dump_list = []
+        for i, obj in enumerate(args):
+            dump_name = 'ObjDump{}'.format(i)
+            self.app.bind(dump_name, obj)
+            dump_list.append(dump_name)
+        self.app.bind('ObjDumpList', dump_list)
         raise DumpException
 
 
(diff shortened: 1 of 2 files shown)


josephmancuso:
When i test this I am only getting "Undefined"

Screen Shot 2020-07-06 at 6 36 32 PM

tpow:
Hmm... I'm not sure what's wrong. I've tested this from within a controller and with dd in a view template.

For example, I changed TestController@show like this:

```python
    def show(self):                                                              
        dd("hello, dd", self)                                                    
        return 'show'
```

and get output something like this:
```
Showing 2 objects:
str hello, dd
- __doc__ - str str(object='') -> str str(bytes_or_buffer[, encoding[, errors]]) -> str Create a new string object from the given object. If encoding or errors is specified, then the object must expose a data buffer that will be decoded using the given encoding and error handler. Otherwise, returns the result of object.__str__() (if defined) or repr(object). encoding defaults to sys.getdefaultencoding(). errors defaults to 'strict'.
```
- - -
```
TestController <app.http.controllers.TestController.TestController object at 0x106de4f60>
- __dict__ - dict : 1 
test: True
- __module__ - str app.http.controllers.TestController
+ test - bool True
```

tpow:
@josephmancuso Any other ideas? I'm not sure why this is failing for you.

josephmancuso:
@tpow not sure. I didn't dig too far into it. Will take another stab tomorrow. If it works for you it should just be something silly for me

josephmancuso:
This works

PR #326, opened 2020-07-23
Next Major Version (3.0)

josephmancuso:
### Changed

- [x] Changed craft key command to put into env by default
- [x] Changed setting of response headers for the response and request headers for the request.
- [x] Changed Request class is no longer accessible before a request is started and is no longer a singleton
- [x] Changed Response status is now done through the response class
- [x] Adds `queue`, `available_at` and `reserved_at` to queue_jobs table.
- [x] Bumps 1 job per queue work poll to 5 jobs per poll

### Added

- [x] Added single method controllers via `__call__` method
- [x] Added a `HeaderBag` on the request class.
- [x] Added `Header` class concept.
- [x] Added same header concept to responses
- [x] Added Vue3 Preset
- [x] Added Secure cookies via a SECURE_COOKIES environment variable

### Fixed

Nothing to see yet

### Removed

- [x] Removed "HTTP" concept on the request class.
- [ ] Removed route helpers


tests/commands/test_new.py
```diff
@@ -0,0 +1,236 @@
...
+        self.command_tester.execute([('target', 'new_project')])
+        self.assertTrue(
+            "Project Created Successfully" in self.command_tester.get_display()
+        )
+        # verify that project has really been created by checking files
+        self.assertTrue("craft" in os.listdir(self.test_project_dir))
+        self.assertTrue("app" in os.listdir(self.test_project_dir))
+
+    @responses.activate
+    def test_can_craft_default_repo_successfully_with_release(self):
```
josephmancuso, line 171: @girardinsamuel can you take a look at this? All these tests are failing

girardinsamuel: Yes I am looking into it

girardinsamuel: On the PR branch (feat/342), all tests are still passing in my local env 
image

I will check difference in requirements

girardinsamuel: It's due to the upgrade in `cleo` version, I might have to edit the tests, but it should not be too complicated. I'll do this later today 😉

girardinsamuel: @josephmancuso 
It's fixed 💪  in https://github.com/MasoniteFramework/masonite/pull/367
The API of cleo commands testers and the Command API changed a bit failing the tests. I had to looked up `cleo` code as the online doc has not been updated since 0.6.2..

I made a second commit to fix some other commands which are not tested thus not failing but which should fail now because of the change in Command cleo API: `self.error` => `self.line_error`.

Issue #339, opened 2020-09-24
Add a way to edit configuration


girardinsamuel:
**Is your feature request related to a problem? Please describe.**
I would like to be able to update configuration of a package in unit tests when testing it.
I would like to update one or more keys of the configuration of a package.

It could be handy to be able to do it outside of tests too.

**What do we currently have to do now?** 
Now we can **get** config with `config()` helper but not **set** it.

**Describe the solution you'd like**
```python
from masonite.helpers import set_config

set_config("application.name", "Brand new name !")
```
Of course, first arg of `set_config` would accept same strings as `config`.

**Describe alternatives you've considered**
- We can change environment loaded during tests, but changing 
env vars during a unit test does not change the config and the `config()` function still returns the same value.

- [ ] Is this a breaking change? No

girardinsamuel:
We have to reimplement the way the configuration is working, to be more solid and powerful, but in the meantime I found a neat way to "update" settings in unit tests with mocking.

```python
from unittest.mock import patch

@patch('config.application.APP_NAME', "Masonite 4.0")
@patch.dict('config.broadcast.DRIVERS', {"pusher": {"app_id": .., "client": None, ...} })
def my_test(self):
     pass
```

girardinsamuel:
This is a nice to have feature especially for testing but it's not required for masonite 3 so moved it out of the project

girardinsamuel:
This won't be done in Masonite 3 and needs some thoughts. I am closing it for now.

PR #341, opened 2020-09-26
Add possibility to set configuration

girardinsamuel:
Add a helper `set_config("application.name", "A new name")`.
More info in #339. 

For that config needs to be stored and not only fetched from env.
As in Laravel, here a new "Config" repository has been created

WORK IN PROGRESS

### DIFF
--- a/src/masonite/config.py
+++ b/src/masonite/config.py
@@ -1,0 +1,22 @@
+
+
+class ConfigRepository(object):
+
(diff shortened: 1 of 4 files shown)


girardinsamuel:
I guess I have to use ConfigParser.
It solves exactly what I am struggling with.

https://docs.python.org/3/library/configparser.html

Issue #342, opened 2020-09-26
Add a way to craft project from repo


girardinsamuel:
[Initially proposed by @josephmancuso], I am opening an issue for it.

**Is your feature request related to a problem? Please describe.**
Add a way to craft project from repositories to have some cool presets

**What do we currently have to do now?** 
Not possible

**Describe the solution you'd like**
```
craft new --repo https://github.com/girardinsamuel/masonite-vue-inertiajs
```
or even
```
craft new --repo girardinsamuel/masonite-vue-inertiajs
```
Maybe adding git provider in arg ?
```
craft new --repo username/repo --provider gitlab (default is github)
```

I'll work on it 💪  !

PR #343, opened 2020-09-29
Add crafting project from Git repository

girardinsamuel:
Implements #342.
This will allow to start project for custom presets/template Masonite projects that we/or the community can maintain.

### DIFF
--- a/src/masonite/commands/NewCommand.py
+++ b/src/masonite/commands/NewCommand.py
@@ -18,7 +18,6 @@ class NewCommand(Command):
     Creates a new Masonite project
 
     new
-        {name? : Name of your Masonite project}
         {target? : Path of you Masonite project}
         {--b|--branch=False : Specify which branch from the Masonite repo you would like to install}
         {--r|--release=False : Specify which version of Masonite you would like to install}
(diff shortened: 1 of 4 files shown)


src/masonite/commands/NewCommand.py
```diff
@@ -14,8 +14,12 @@ class NewCommand(Command):
         {target? : Path of you Masonite project}
         {--b|--branch=False : Specify which branch from the Masonite repo you would like to install}
         {--r|--release=False : Specify which version of Masonite you would like to install}
+        {--rp|--repo=False : Specify from which repository you want to craft your: girardinsamuel/masonite-base-project}
```
josephmancuso, line 17: it would be better he instead of doing =False just specify our cookie-cutter repo that it installs from now

josephmancuso (changes requested):

src/masonite/commands/NewCommand.py
girardinsamuel: Yes I agree 😉

girardinsamuel:
- I finally refactored a bit the command to allow more modularity to change providers.
- I also added unit tests for commands, so I have experimented with `cleo` testing helpers => really useful, we will be able to test other commands...one day 😉 

- [ ]  add one test for GitLab
- [ ]  argument name was not used before I started to use the PR: I can either remove it totally or start using it and concatenate the target dir to create a folder named after `name`, but I guess it could be confusing, I would like to let things 
simple

josephmancuso (approved):

Issue #128, opened 2018-11-12
[Docs] Acknowledge Laravel inspiration


superbiche:
Hey there,

I found this framework on Dev community, coming from a Laravel background and having to experiment on Python for a client (I need to migrate a legacy app), this framework seems to be exactly what I need.

You say in #62 that Laravel inspired some parts of the framework, which is great - Laravel is really amazing to develop with - but I think acknowledging this on the README or on the homepage (or both) would be nice :)

Also, this might help some fellow Laravel developers looking for a Python framework with the same DX as Laravel to choose Masonite instead of any other framework.

Cheers and thanks for the work

josephmancuso:
Not a bad idea. You can open a PR here if you like https://github.com/MasoniteFramework/docs

Might be a good idea on changing where you think the best place would be for it.

PR #346, opened 2020-10-02
Add with_success and with_flash request helper

girardinsamuel:
As in Laravel this PR adds in addition to `with_errors`:
* with_success
* with_flash (base helper)

So we can do :
```python
request.redirect('users').with_success('User created !')
request.redirect('users').with_flash('success', 'User created !')
request.redirect('users').with_flash('key', val)
```

I added one test for `with_errors` as it was not tested and tests for the two new helpers.
The documentation is up to date [PR #128](https://github.com/MasoniteFramework/docs/pull/128).


girardinsamuel:
Same for this small helpers which are not breaking changes

josephmancuso (approved):

PR #348, opened 2020-10-08
Fix issue with namespace which effects when installing one of the pac…

ar45:
…kages externally

https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages

### DIFF
--- a/src/masonite/__init__.py
+++ b/src/masonite/__init__.py
@@ -1,3 +1,7 @@
+from pkgutil import extend_path
+__path__ = extend_path(__path__, __name__)
+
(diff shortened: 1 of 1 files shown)


girardinsamuel:
Hi @ar45 Masonite is using Native namespace packages for now I guess
https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages
As Masonite only supports Python 3+, I don't see any reason to change the packaging style but I might be wrong

What is the issue you encounter precisely ? 🤔

PR #360, opened 2020-10-28
Fix/handlers

josephmancuso:
Screen Shot 2020-10-27 at 10 22 47 PM


josephmancuso:
Adds a a better debugging screen for application/json requests in debug mode

PR #373, opened 2020-12-05
fix/372

### LINKED ISSUE #372: Wrong http status code
**Describe the bug**
When sending a POST request to the GET route, it should display "Method Not Allowed", but now it displays "Not Found".

**To Reproduce**
1. craft new testproj
2. craft serve
3. curl -d "" http://127.0.0.1:8000/

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: Win10

imlison:
Fixes #372 : Return 405 when the request method is incorrect


josephmancuso (approved):
Thanks!

PR #385, opened 2021-01-12
routing rewrite

josephmancuso:
This is an experimental feature started to discover the effort required to rewrite the routing of Masonite.

Routing is another feature that is one of the oldest that have not made it past a rewrite. I do not know if this will make it into the Masonite 3 release. Possibly Masonite 4.

This changes the current routing structure which is something like this:

```python
from masonite.routes import Get, Post, Put

ROUTES = [
    Get('/url', 'Controller@show'),
    Post('/login', 'Controller@show'),
    Put('/update', 'Controller@show')
]
```

to this:

```python
from masonite.routes import Route

Route.get('/url', 'Controller@show')
Route.post('/login', 'Controller@show')
Route.put('/update', 'Controller@show')
```

## TODO:

- [ ] Remove tests relating to old route style
- [ ] Add domains to routing
- [ ] Update / Fix TODOs in the PR code
- [ ] Remove Sanity code
- [ ] Add request params to the controller methods
- [x] Update web.py file to new route style
- [x] Update RouteProvider for new route style
- [ ] Update auth routes to use new style of routing


josephmancuso:
Route URLs and compilers will look the same:

```
/dashboard/@id/@user_id:int
```

josephmancuso:
Route groups will be changed slightly. 

We will go from this format:

```python
ROUTES = [
    RouteGroup([
        Get('/url', 'Controller@show'),
        Post('/login', 'Controller@show'),
    ], prefix="/dashboard")
]
```

To this:

```python
Route.group(
    Route.get('/url', 'Controller@show'),
    Route.post('/login', 'Controller@show'),
    prefix="/dashboard"
)
```

josephmancuso:
Middleware will not change.

```python
Route.get('/url', 'Controller@show').middleware(["auth"]),
```

josephmancuso:
route naming also remains the same:

```python
Route.get('/url', 'Controller@show').name("dashboard.index"),
```

josephmancuso:
@Marlysson @girardinsamuel I'd like your thoughts on this as well.

I do plan to not put this in Masonite 3 as this is sort of a big change for people. It would maybe only take an hour at most to upgrade people routing in their apps but it is just another step.

On the other hand I would like to upgrade all the routing into something more modern. Right now I feel like the routing is a little hacky .. It is what I would call legacy code at this point so it is time for a rewrite for it

girardinsamuel:
This API looks great. It might be a good thing not to release it in the same time as ORM.

About resource ? Should we also do this ?
```python
Route.resource('/url', 'Controller')
```

josephmancuso:
Yeah exactly. We're basically going to start deprecating the old class based routing and add this new style.

Marlysson:
I loved the new syntax and turn into something more direct. Tied more the methods indicating that they are Routes.

jpmateo022:
I like it, it gives me a Laravel vibe, and it's much clearer. It makes us no need to import GET, POST, and ... individually.

PR #389, opened 2021-01-15
add missing with_errors, with_success in 3.0

girardinsamuel:
- add helpers released in 2.X and missing in 3.0
- add tests coming along those helpers

it's basically the same PR just with different target


josephmancuso (approved):

josephmancuso:
I merged 2.3 into develop so the code is there now

PR #401, opened 2021-01-21
Add view testing assertions

### LINKED ISSUE #257: Testing: assertViewIs assertion
**Is your feature request related to a problem? Please describe.**
It would be nice if we could assert the name of the view that is returned.
```python
self.assertTrue(self.get('/login').assertViewIs('auth.login'))
```

girardinsamuel:
Small PR to add following assertions helpers to test views:
- assertViewIs
- assertViewHas
- assertViewHasAll
- assertViewMissing
- assertRedirect

Here the assertions are used to test data made available in view context and view template name.

Note: this is not what has been proposed in #257. What has been proposed in this issue is already implemented and is named `assertIsNamed`. I will close #257 then.

Ready for review ;)


girardinsamuel:
@josephmancuso  you just merged doc for those new view assertions but the assertions has not been released yet 😉 . But this is ready so if you have time to review it that's nice ! 

No rocket science stuff here lol just some basic asserts, I just had to save the view instance to be able to assert it later (through `original` attribute)

PR #403, opened 2021-01-25
fix Problems when not in GMT time zone

puzzle9:
i’m from GMT + 8:00
When I run 'queues' and' wait '
Generated ` available_ It's always wrong
So I tried to modify this part of the code


puzzle9:
Maybe there should be some other solution
But this problem has been bothering me for a long time

PR #411, opened 2021-01-31
Add easy way for building packages for masonite

girardinsamuel:
[ON HOLD] This PR is on hold, waiting for the refactoring of routes. Else it completed at 80 % I would say 💪 

Add a `PackageServiceProvider` that you can use in your packages, to easily:
- publish a package config file
- register routes (and publish them if you want to override them)
- register views (and publish them if you want to override them)
- register migrations (and publish them if you want to override them)
- include assets to serve as static files (and publish them if you want to override them)
- register commands

All files are published under a configurable "vendor/package-name" namespace.
This is how your package provider would look like:

```python
from masonite.providers import PackageProvider

package_root_path = "path/to/package/root"

class MyPackageProvider(PackageProvider):

    def configure(self):
        self.name("my-package")
        self.add_config("test")
        self.add_command(MyCommand())
        self.add_routes("admin", "public", tag="my-routes")
        self.add_migrations("create_my_table", "create_my_other_package_table", tag="db")
        self.add_assets(
                {
                    "package.js": "app.js",
                    "package.css": None,
                    "sub/admin.js": "admin.js",
                }
       )
       self.add_command(OtherCommand(), "DemoCommand")
```

Of course you can still use existing methods and capabilities of the base ServiceProvider to register and publish stuff from your package. But this should really save you time and enforce a good package structure ! 🚀 

This will come with parameters in some of the methods to above to override publishing tags, override default folders in your package ...

Finally using this package provider can provide( if enabled) a help command to your package to show:
- what can be published in your project
- what is registered by the package when installed

---
The idea is based on the package made for Laravel
https://github.com/spatie/laravel-package-tools

### DIFF
--- a/src/masonite/commands/PublishCommand.py
+++ b/src/masonite/commands/PublishCommand.py
@@ -19,9 +19,9 @@ class PublishCommand(Command):
                 if self.option("tag") != "None":
                     provider.publish(tag=self.option("tag"))
                     provider.publish_migrations(tag=self.option("tag"))
-
-                provider.publish()
-                provider.publish_migrations()
+                else:
+                    provider.publish()
+                    provider.publish_migrations()
 
                 return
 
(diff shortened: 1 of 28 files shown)


girardinsamuel:
This will be usred/integrated with Masonite 4 finally.
To be migrated into M4.

PR #424, opened 2021-03-05
Add missing log query config in database.py

erhuabushuo:
add log_queries config, otherwise https://orm.masoniteproject.com/installation#logging is not working


girardinsamuel:
The changes you made should also be done in newly scaffolded Masonite project (when using `python craft new`). For this you could open a PR on the cookiecutter repo: https://github.com/MasoniteFramework/cookie-cutter on 3.0 branch. 

Thanks 😉

PR #449, opened 2021-07-22
Fixes Issue Headers issue in TestCase

yubarajshrestha:
Behaviour of `withHeaders()`  helper was disabled in test cases.

This patch will fix issues with withHeaders test cases.


josephmancuso:
This looks good. You can ignore those tests although you can fix the linting error by running `make format` in the project and commit those changes

yubarajshrestha:
> This looks good. You can ignore those tests although you can fix the linting error by running `make format` in the project and commit those changes

Done. that csrf test failure exists in my pc as well. I tested with that test commented everything works perfect.

src/masonite/testing/MockRoute.py
```diff
@@ -379,3 +379,17 @@ def assertRedirect(self, redirect_uri):
         self.route.get_response()
         assert response.is_status(302) or response.is_status(301)
         assert request.redirect_url == redirect_uri
+
+    def assertDatabaseHas(self, schema, value):
```
josephmancuso, line 383: These should really go in the TestCase class and not the MockRoute class

josephmancuso (changes requested):

src/masonite/testing/MockRoute.py
yubarajshrestha: Actually that part is already in TestCase class but the above solution is for different issue. 

**Case 1**
```python
self.post("/your-api-endpoint", {
            "name": "Awesome",
}).assertDatabaseHas("table_name.column_name", "Awesome")
```

**Case 2**
```python
self.post("/your-api-endpoint", {
            "name": "Awesome",
})

self.assertDatabaseHas("table_name.column_name", "Awesome")
```

In above samples, the **Case 2** works perfectly fine because `assertDatabaseHas` already exists in TestCase, but if you try **Case 1** it gives error saying `assertDatabase is not available in MockRoute`. That's why I added those.

yubarajshrestha: Following is the exact error message you'll get in **Case 1**. 
`AttributeError: 'MockRoute' object has no attribute 'assertDatabaseHas'`

yubarajshrestha: @josephmancuso any reviews on this?

josephmancuso:
@girardinsamuel this is fixed in M4 right?

PR #462, opened 2021-12-23
do not need decode to str when deserialize json to Python object

erhuabushuo:
[https://github.com/python/cpython/blob/71ef0b4c2b77195bb1adc42602549284f7ee9566/Lib/json/__init__.py#L301](https://github.com/python/cpython/blob/71ef0b4c2b77195bb1adc42602549284f7ee9566/Lib/json/__init__.py#L301)

"""Deserialize ``s`` (a ``str``, ``bytes`` or ``bytearray`` instance
    containing a JSON document) to a Python object.

json.loads may accept  str, bytes, bytearray, we don't need decode to str

### DIFF
--- a/src/masonite/drivers/authentication/AuthJwtDriver.py
+++ b/src/masonite/drivers/authentication/AuthJwtDriver.py
@@ -9,7 +9,6 @@ from ...helpers import config, cookie_expire_time
 from ...app import App
 
 
-
 class AuthJwtDriver(BaseDriver, AuthContract):
     def __init__(self, app: App):
         """AuthCookieDriver initializer.
--- a/src/masonite/request.py
+++ b/src/masonite/request.py
@@ -309,9 +309,6 @@ class Request(Extendable):
 
             request_body = self.environ["wsgi.input"].read(request_body_size)
 
-            if isinstance(request_body, bytes):
-                request_body = request_body.decode("utf-8")
-
             return json.loads(request_body or "{}")
         else:
             fields = cgi.FieldStorage(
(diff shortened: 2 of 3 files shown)


erhuabushuo:
_get_standardized_value is not working when json payload contains bool, float and None type.

```
if isinstance(value, (str, int, dict)):
        return value
```
json may contains bool, float, and None type.

josephmancuso:
I'll have to manually

erhuabushuo:
wow, I didn't modify src/masonite/drivers/authentication/AuthJwtDriver.py:13:1: E303 too many blank lines (3)

erhuabushuo:
lint code

josephmancuso:
@erhuabushuo can I get a little explanation here? where did this issue come from and how to I test the use case which its broken for? I don't like merging in code I don't understand

josephmancuso:
@girardinsamuel this should be good to merge right? Code looks good and tests pass though im still kind of confused on what the issue originally was

girardinsamuel:
> @girardinsamuel this should be good to merge right? Code looks good and tests pass though im still kind of confused on what the issue originally was

Let me check locally first.

erhuabushuo:
POST_DATA may be "application/json" content, the json content may contains float, bool and None .

currently if json contain these type (bool, float, None) will be raised exception,

```
if isinstance(value, (str, int, dict)):  # here can't be true
        return value


if not value.filename: # here will be raise exception, cause bool , float and None don't have the property of filename .
        return value.value
```

PR #468, opened 2021-12-24
Feature/465 - Adds API module back into Masonite

josephmancuso:


src/masonite/commands/MakeControllerCommand.py
```diff
@@ -35,7 +35,7 @@ def handle(self):
         content = render_stub_file(stub_path, name)
 
         filename = f"{name}.py"
-        path = controllers_path(filename)
+        path = config_path(filename)
```
girardinsamuel: Why did you rename this 🤔 ?

josephmancuso: was a mistake lol

src/masonite/api/Api.py
```diff
@@ -0,0 +1,93 @@
...
+        return self
+
+    def generate_token(self):
+        secret = self.config.get("jwt").get("secret")
+        algorithm = self.config.get("jwt").get("algorithm")
+        expire_minutes = self.config.get("jwt").get("expires")
+        version = self.config.get("jwt").get("version")
+        if expire_minutes:
+            expire_minutes = (
+                pendulum.now(tz="GMT").add(minutes=expire_minutes).to_datetime_string()
```
girardinsamuel: Should we use `GMT` or `UTC`here ?

src/masonite/api/commands/APIInstallCommand.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __init__(self, application):
+        super().__init__()
+        self.app = application
+
+    def handle(self):
+        stub_path = self.get_stub_config_path()
+
+        content = render_stub_file(stub_path, "api.py")
+        print(content)
```
girardinsamuel: a little print to remove..

src/masonite/api/commands/APIInstallCommand.py
```diff
@@ -0,0 +1,57 @@
...
+
+        path = config_path("api.py")
```
girardinsamuel, line 28: This block looks duplicated !

tests/features/api/test_api.py
```diff
@@ -0,0 +1,17 @@
+from tests import TestCase
+from src.masonite.api.facades import Api
+from src.masonite.headers import Header
+
+
+class TestApiModule(TestCase):
+    def test_can_can_token_from_input(self):
```
girardinsamuel:
```suggestion
    def test_can_get_token_from_input(self):
```

tests/features/api/test_api.py
```diff
@@ -0,0 +1,17 @@
...
+        request = self.make_request()
+        request.input_bag.add_post("token", "secret")
+
+        self.assertEqual(Api.get_token(), "secret")
+
+    def test_can_can_token_from_input_from_header(self):
```
girardinsamuel:
```suggestion
    def test_can_get_token_from_header(self):
```

src/masonite/input/InputBag.py
```diff
@@ -187,3 +187,7 @@ def parse_dict(self, dictionary):
                 new_dict.update({name: value})
 
         return new_dict
+
+    def add_post(self, key, value):
```
girardinsamuel, line 191: Where this method is needed ?

girardinsamuel (changes requested):
Here is a quick review !

src/masonite/api/Api.py
josephmancuso: There is actually no difference. but to make it consistent ill change it to UTC. We use UTC in the masonite-orm repo for example. Good catch

src/masonite/api/commands/APIInstallCommand.py
josephmancuso: good catch

src/masonite/api/commands/APIInstallCommand.py
josephmancuso: good catch

src/masonite/input/InputBag.py
josephmancuso: I believe I use it in one of the tests

src/masonite/api/Api.py
girardinsamuel: oh okay ! I was thinking it was different I'll go check those terms then 😅

src/masonite/input/InputBag.py
girardinsamuel: oh okay yes makes sense

src/masonite/api/Api.py
```diff
@@ -0,0 +1,93 @@
...
+        if not expires:
+            return True
+
+        expired = pendulum.parse(expires, tz="UTC").is_past()
+
+        if expired:
+            return False
+
+        if version:
+            if unencrypted_token["version"] != version:
```
josephmancuso, line 60: I just realized this check has to happen first or else it will never get here if the expiration is set to None. even if tokens never expire we still need to verify token version

josephmancuso:
Fixed JWT Guard

josephmancuso:
Also fixed a big issue where routes weren't even aloud to have multiple middleware stacks assigned to them

josephmancuso (commented):
reviewed myself. Why can't we approve our own PR's?

PR #480, opened 2022-01-05
Add a way to register new presets easily

girardinsamuel:
A small PR to easily register new presets with Masonite 4 philosophy

- add provider for presets to allow registering new presets
- fix creating dirs if they don't exist
- add a `presets()` method in PackageProvider, to be able to register presets easily from packages. Of course we can also register a preset just by doing `app.make("presets").add(MySuperPreset())`.
- add a `--list` flag to preset command to be able to easily list available presets.


josephmancuso (approved):
This is a breaking change right?

josephmancuso:
Since the preset provider is not in new projects or projects created with M4 this would actually remove the preset options from existing projects

girardinsamuel:
> Since the preset provider is not in new projects or projects created with M4 this would actually remove the preset options from existing projects

@josephmancuso 
Hmm you're right ! Should we make temporarily this binding in an existing provider. and at next major release remove the binding and update project template to add the new provider ?

src/masonite/providers/FrameworkProvider.py
```diff
@@ -1,22 +1,35 @@
...
 
-class FrameworkProvider:
+
+class FrameworkProvider(Provider):
     def __init__(self, application):
         self.application = application
 
     def register(self):
-        pass
+        # @M5 remove this and add PresetsProvider in default project
```
girardinsamuel, line 17: For now I put PresetsProvider code here. @josephmancuso does this work for you or do you have a better idea ?

josephmancuso (approved):

PR #492, opened 2022-01-12
Ensure that hashing algorithms return string

### LINKED ISSUE #491: Hash.make should default to a string and not bytes
could also have a hash.make_bytes() if we need current behavior

Marlysson:
@josephmancuso Closes #491

### DIFF
--- a/tests/features/hashing/test_hashers.py
+++ b/tests/features/hashing/test_hashers.py
@@ -34,7 +34,7 @@ class TestHashers(TestCase):
 
     def test_should_return_hash_content_as_string(self):
 
-        drivers = ["argon2", "bcrypt"]
+        drivers = Hash.drivers.keys()
 
         for driver in drivers:
             with self.subTest(driver=driver):
--- a/tests/features/hashing/test_hashers.py
+++ b/tests/features/hashing/test_hashers.py
@@ -33,7 +33,7 @@ class TestHashers(TestCase):
         assert Hash.needs_rehash(hashed, driver="argon2")
 
     def test_should_return_hash_content_as_string(self):
-      
+
         drivers = ["argon2", "bcrypt"]
 
         for driver in drivers:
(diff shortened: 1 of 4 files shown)


tests/features/hashing/test_hashers.py
```diff
@@ -31,3 +31,12 @@ def test_argon2_needs_rehash(self):
         # Here argon2 method is invoked without custom options and will
         # use default argon2 configuration, so rehash will be needed
         assert Hash.needs_rehash(hashed, driver="argon2")
+
+    def test_should_return_hash_content_as_string(self):
+
+        drivers = ["argon2", "bcrypt"]
```
girardinsamuel: To avoid hard-coding drivers list, you could use this:
```suggestion
        drivers = Hash.drivers.keys()
```
(It's equivalent to `self.application.make("hash").drivers.keys()`

tests/features/hashing/test_hashers.py
```diff
@@ -31,3 +31,12 @@ def test_argon2_needs_rehash(self):
...
+
+        for driver in drivers:
+            with self.subTest(driver=driver):
```
girardinsamuel: Here what don't you just do ?
```python
        for driver in drivers:
            hashed = Hash.make("masonite", driver=driver)
            assert type(hashed) == str
```
🤔 Did I miss something ?

girardinsamuel (changes requested):

tests/features/hashing/test_hashers.py
Marlysson: With subTest ( and using driver as parameter ) let us know which driver fail during run.

tests/features/hashing/test_hashers.py
Marlysson: Great!!

tests/features/hashing/test_hashers.py
girardinsamuel: oh okay !

girardinsamuel (approved):

josephmancuso (approved):

PR #497, opened 2022-01-18
Fix/496 - Fix controller's name when in creation by craft controller there the slash between folder and controller's name.

### LINKED ISSUE #496: Controller creation with slash in its name like "craft controller api/NameController" creates invalid name
**Describe the bug**
When run creating a controller named with slash(/) ( indication inside folder in creation ) then the name of controller is append the slash, instead just last part after slash (/)

**To Reproduce**
1. python craft controller api/PodcastController
2. Folder named as "api" is created and controller name it's **class Api/PodcastController(Controller)**

**Expected behavior**
1. Create the folder named as "api"
2. Create controller named as "PodcastController" inside "api" folder ( or what else folder is defined )

**Desktop:**
 - OS: Windows 10
 - Version Masonite 4.2.0

Marlysson:
Closes #496 
Fix controller's name when in creation by craft controller there the slash between folder and controller's name.

I've tried handle relative paths with controllers_path() from better way 😅 
Maybe needed a refactor just let me know.

@josephmancuso I would implement tests, but I didn't found any tests for commands for now in project.


josephmancuso:
We should be able to do this in a cleaner way ill have to take a look. This is a little messy tbh.

josephmancuso (approved):

PR #501, opened 2022-01-20
Fix publishing views

### LINKED ISSUE #493: Loading views with namespace should look in packages vendor
Right now loading views from a namespace will look only in corresponding Python namespace meaning that it can load view from a internal project namespace or directly from a package.

What's missing is loading view from package vendor directory if the user has published the view, it should be done first to be able to override package views.

It should be done in `views()` of PackageProviders.

Also all directories registered from there should be used, now there is a `views[0]` which look wrong.

girardinsamuel:
Fixes #493.

- [x] PackageProvider `views()` method now accepts only one folder. This folder should contain all the views you want to be able to use / publish when the package is installed.
- [x] Removed a duplicated method in View
- [x] Fixed publishing only .html in views folder to avoid this:
image

### DIFF
--- a/src/masonite/packages/Package.py
+++ b/src/masonite/packages/Package.py
@@ -32,9 +32,13 @@ class Package:
         self.config = self._build_module_path(config_path)
         return self
 
-    def add_views(self, *locations):
-        for location in locations:
-            self.views.append(self._build_module_path(location))
+    def add_views(self, location):
+        views_folder = self._build_module_path(location)
+        if not os.path.isdir(views_folder):
+            raise ValueError(
+                "views() first argument must be a folder containing all package views."
+            )
+        self.views = views_folder
         return self
 
     def add_migrations(self, *migrations):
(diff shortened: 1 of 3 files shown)


yubarajshrestha:
@girardinsamuel looks like you are publishing contents into vendor directory. So does it mean we have to use : `view.render('vendor.instant_article.feeds')`?

girardinsamuel:
> @girardinsamuel looks like you are publishing contents into vendor directory. So does it mean we have to use : `view.render('vendor.instant_article.feeds')`?

No, package provider takes care of registering those views with a namespace. It means that you can use `view.render("instant_article:feeds")`.

It also means that, when views are not published they are taken directly from package folder and if they are published, then views are taken from project vendor directory. This allow users to override package views by publishing them 😉

yubarajshrestha:
> > @girardinsamuel looks like you are publishing contents into vendor directory. So does it mean we have to use : `view.render('vendor.instant_article.feeds')`?
> 
> No, package provider takes care of registering those views with a namespace. It means that you can use `view.render("instant_article:feeds")`.
> 
> It also means that, when views are not published they are taken directly from package folder and if they are published, then views are taken from project vendor directory. This allow users to override package views by publishing them 😉

Wow, just like I wanted. It's really great. Thank you so much.

josephmancuso (approved):

PR #504, opened 2022-01-21
Add Azure storage driver

girardinsamuel:
All is working except `get_files()` method and I don't know why..


josephmancuso:
These are still WIP. Are these good ? Or is the get_files not working? If thats not working whats wrong with it? I can take a look then

PR #506, opened 2022-01-21
 Throw exception if route is not found when in debug mode and add http error views 

### LINKED ISSUE #487: Throw exception when in debug mode if route is not found

### LINKED ISSUE #512: Add missing debug mode
In Masonite 3 we had `APP_DEBUG` env variable which could configure if debug mode was enabled or not.

- [ ] this should be added back as a parameter into `config/application.py`
- [ ] a helper `is_debug()` should be added in `application`
- [ ] the mode should be used to display exception page or error 500 page in case of error

Also how we use this with `APP_ENV` ? Should it be related ? Can we have environment equal to `production` and debug enabled ?

girardinsamuel:
### Add debug mode (fix #512)
 - controlled by `APP_ENV` env variable and defined in `config/application.py` so it can be accessed with `config("application.debug")`
 - a helper is available: `application.is_debug()`

### When debug mode is on
all exceptions are rendered with exceptionite (except for application/json) fix #487.
image

### When debug mode is off
- all exceptions which are HTTP errors (500, 404, 403) are rendered with a view located in `errors/{code}.html`. If view does not exist the message is rendered directly as a string with the correct status.
- all exceptions which have a `is_http_exception` flag will be rendered like above.
- any other exception which does not fit here will be rendered as a 500 exception with the page (because when debug is not enable we dont want to see exceptionite page).

### Customization
- HTTP error views can be overriden by overriding .html files
- For more possibility, you can override the `HTTPExceptionHandler` to do what you want.

### Tests helpers
- add `debugMode` context manager to change debug mode locally during a test
- add `env` context manager to change env locally during a test

### Tinker command
- now displays environment, debug mode and masonite version for simplicity:
image


src/masonite/tests/TestCase.py
```diff
@@ -152,7 +155,15 @@ def fetch(self, path, data=None, method=None):
             return self.application.make("tests.response").build(
                 self.application, request, response, route
             )
-        raise Exception(f"No route found for url {path}")
+
+        exception = RouteNotFoundException(f"No route found for url {path}")
+        if self._exception_handling:
+            response = self.application.make("exception_handler").handle(exception)
```
girardinsamuel, line 161: This way when exception handling is enabled but route not found during tests we can use assertions !

tests/core/authorization/test_gate.py
```diff
@@ -102,19 +102,22 @@ def test_gate_has_permission(self):
...
 
     def test_authorize(self):
+        self.withExceptionsHandling()
         self.gate.define("display-admin", lambda user: False)
-        with self.assertRaises(AuthorizationException) as e:
-            self.gate.authorize("display-admin")
-            exception = e.exception
-        self.assertEqual(e.exception.message, "Action not authorized")
-        self.assertEqual(e.exception.status, 403)
+        with self.debugMode():
```
girardinsamuel, line 107: in debug mode we can see the exception and assert it

tests/core/authorization/test_gate.py
```diff
@@ -102,19 +102,22 @@ def test_gate_has_permission(self):
...
+            self.assertEqual(e.exception.message, "Action not authorized")
+            self.assertEqual(e.exception.status, 403)
 
     def test_authorize_in_controller(self):
         self.withExceptionsHandling()  # this will allow exception to be handled and rendered
         self.gate.define("display-admin", lambda user: False)
-        self.get("/not-authorized").assertForbidden().assertContains(
-            "Action not authorized"
-        )
+        with self.debugMode(False):
```
girardinsamuel, line 117: if debug mode is off, the view rendered is the 403 error page

tests/integrations/config/application.py
```diff
@@ -1,14 +1,15 @@
-import os
+from masonite.environment import env
 
-KEY = os.getenv("APP_KEY", "-RkDOqXojJIlsF_I8wWiUq_KRZ0PtGWTOZ676u5HtLg=")
+KEY = env("APP_KEY", "-RkDOqXojJIlsF_I8wWiUq_KRZ0PtGWTOZ676u5HtLg=")
 
+DEBUG = env("APP_DEBUG", True)
```
girardinsamuel, line 5: here for existing projects without this parameter it won't raise error.
`config("application.debug")` and `application.is_debug()` will be `False`

girardinsamuel:
@josephmancuso here is my proposal. ready to discuss it.

src/masonite/foundation/Application.py
```diff
@@ -48,9 +49,16 @@ def get_providers(self):
     def __call__(self, *args, **kwargs):
         return self.response_handler(*args, **kwargs)
 
+    def is_debug(self):
+        """Check if debug mode is enabled."""
+        return bool(config("application.debug"))
```
josephmancuso: The problem is this won't work for existing applications. The new 4.0 apps we didn't add this config variable in the config file

src/masonite/providers/RouteProvider.py
```diff
@@ -59,7 +60,9 @@ def boot(self):
                     )
 
             else:
-                response.view("route not found", status=404)
+                raise RouteNotFoundException(
```
josephmancuso, line 63: Doesn't this need to check for debug mode?

tests/integrations/config/application.py
josephmancuso: if it will always be false for existing projects, how do they enable it?

josephmancuso (changes requested):

src/masonite/providers/RouteProvider.py
girardinsamuel: No that's the role of the exception handler. Depending of the exceptions and the debug mode it will handle this differntly

src/masonite/foundation/Application.py
girardinsamuel: It will work. At least, it won't raise an error. this will return False.
So all existing projects won't be in debug mode until they add the line in `config/application.py`.

So it's still a problem I agree, we can until next major release, use:
```python
    def is_debug(self):
        """Check if debug mode is enabled."""
        if Config.has("application.debug"):
             return bool(config("application.debug"))
        else:
             return env("APP_DEBUG")   # using env() here to ensure that it's converted to bool
```

tests/integrations/config/application.py
girardinsamuel: you're right ! Here is what we can do until a major release (maybe): https://github.com/MasoniteFramework/masonite/pull/506/files/f91ba5df468a9fb11a34af5a3913190c4d97788b#r790302614

src/masonite/foundation/Application.py
josephmancuso: That will work yeah

src/masonite/configuration/Configuration.py
```diff
@@ -63,6 +63,9 @@ def merge_with(self, path, external_config):
     def set(self, path, value):
         self._config[path] = value
 
+    def has(self, path):
+        return path in self._config[path]
```
josephmancuso: Is _config a dictionary? Won't this throw an exception if path is not in _config? Shouldn't this be `path in self._config.get(path)`

girardinsamuel: My bad it should be : `return path in self._config`  `_config` is a dotty dict accessible with dots.

girardinsamuel:
I made some updates to the PR it should be good now. We can take many approaches to do this of course... I tried many things and choosed one way.

You can look closely at the updates in ExceptionHandler which is making the switch between debug and not debug and the HttpExceptionHandler which is there to give developers a way to customize things easily (adding call to sentry, or logging stuff or doing anything they want).

josephmancuso:
Tested this with APP_DEBUG false and got this error:

```
  File "/Users/personal/programming/masonite/masonite/src/masonite/exceptions/ExceptionHandler.py", line 54, in handle
    return self.application.make("HttpExceptionHandler").handle(exception)
  File "/Users/personal/programming/masonite/masonite/src/masonite/exceptions/HttpExceptionHandler.py", line 9, in handle
    if self.application.make("view").exists(view_name):
  File "/Users/personal/programming/masonite/masonite/src/masonite/views/view.py", line 150, in exists
    self.load_template(template)
  File "/Users/personal/programming/masonite/masonite/src/masonite/views/view.py", line 224, in load_template
    template_loaders.append(PackageLoader(package_name, package_path))
  File "/Users/personal/Desktop/test 4/venv4/lib/python3.9/site-packages/jinja2/loaders.py", line 319, in __init__
    raise ValueError(
ValueError: The 'templates' package was not installed in a way that PackageLoader understands.
ERROR:waitress:Exception while serving /unkown_route
Traceback (most recent call last):
  File "/Users/personal/programming/masonite/masonite/src/masonite/foundation/response_handler.py", line 27, in response_handler
    application.resolve(provider.boot)
  File "/Users/personal/programming/masonite/masonite/src/masonite/container/container.py", line 172, in resolve
    return obj(*objects)
  File "/Users/personal/programming/masonite/masonite/src/masonite/providers/RouteProvider.py", line 65, in boot
    raise RouteNotFoundException(
masonite.exceptions.exceptions.RouteNotFoundException: GET /unkown_route : 404 Not Found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/personal/Desktop/test 4/venv4/lib/python3.9/site-packages/waitress/channel.py", line 350, in service
    task.service()
  File "/Users/personal/Desktop/test 4/venv4/lib/python3.9/site-packages/waitress/task.py", line 171, in service
    self.execute()
  File "/Users/personal/Desktop/test 4/venv4/lib/python3.9/site-packages/waitress/task.py", line 441, in execute
    app_iter = self.channel.server.application(environ, start_response)
  File "/Users/personal/programming/masonite/masonite/src/masonite/foundation/Application.py", line 51, in __call__
    return self.response_handler(*args, **kwargs)
  File "/Users/personal/Desktop/test 4/venv4/lib/python3.9/site-packages/whitenoise/base.py", line 85, in __call__
    return self.application(environ, start_response)
  File "/Users/personal/programming/masonite/masonite/src/masonite/foundation/response_handler.py", line 29, in response_handler
    application.make("exception_handler").handle(e)
  File "/Users/personal/programming/masonite/masonite/src/masonite/exceptions/ExceptionHandler.py", line 54, in handle
    return self.application.make("HttpExceptionHandler").handle(exception)
  File "/Users/personal/programming/masonite/masonite/src/masonite/exceptions/HttpExceptionHandler.py", line 9, in handle
    if self.application.make("view").exists(view_name):
  File "/Users/personal/programming/masonite/masonite/src/masonite/views/view.py", line 150, in exists
    self.load_template(template)
  File "/Users/personal/programming/masonite/masonite/src/masonite/views/view.py", line 224, in load_template
    template_loaders.append(PackageLoader(package_name, package_path))
  File "/Users/personal/Desktop/test 4/venv4/lib/python3.9/site-packages/jinja2/loaders.py", line 319, in __init__
    raise ValueError(
ValueError: The 'templates' package was not installed in a way that PackageLoader understands.
```

josephmancuso:
I believe this is due to the fact that I don't have a templates/errors/500.html page

josephmancuso:
Everything else seems to work fine

josephmancuso (changes requested):
Just need to fix the assumption that the templates errors directory exists

josephmancuso (changes requested):
Just need to fix the assumption that the templates errors directory exists

girardinsamuel:
> Just need to fix the assumption that the templates errors directory exists

Actually it's more a problem in the view class. The exists() method is not totally right and add_namespaced_location() works for package but does not work for views inside the project itself. We could check that later.
In the meantime, I remove the `errors` namespace.

Now this works when templates exists or not.

girardinsamuel:
I also added default errors pages in cookie-cutter : https://github.com/MasoniteFramework/cookie-cutter/pull/35
😉 
@josephmancuso ready to be merged !

josephmancuso (approved):

PR #509, opened 2022-01-22
Handle `ModelNotFound` exceptions as HTTP 404 errors

### LINKED ISSUE #322: Find_or_404
I'm not sure if this would fit better in the masonite framework or the ORM, so putting it here.

It is often the case in view controllers that you want to do a lookup in a model and either show it in a view or return a 404 not found if the record with that id can't be retrieved. This essentially replaces the find(id) method and would provide a nice shorthand.

I see that the ORM has a find_or_fail method, but at least when using `craft serve` this doesn't seem to return a 404. Instead it raises an exception. It would be helpful to have a method that returns a 404 instead.

girardinsamuel:
Fixes #322.

This PR handles `ModelNotFound` exception raised by ORM `User.find_or_fail` or Query builder `find_or_fail` method. It's handled by the HTTPExceptionHandler to return a 404.

In debug mode we see the exception page and in production we see the 404 error page.


girardinsamuel:
@josephmancuso I fixed the tests here 😉

girardinsamuel:
@josephmancuso 
Do you like this API, what are your thoughts ?

josephmancuso:
@girardinsamuel Hmm i think i prefer if this called an exception inside Masonite ORM and then we just handled whatever exception that throws.

I don't like the idea of importing a special helper just to get a find or fail.

I prefer if this was a method on the model:

```python
User.find(1)
User.find_or_fail(1)
```

we can build an exception in ORM and raise that exception and use Masonite to handle that exception

girardinsamuel:
Okay you're right it might be better !

josephmancuso:
@girardinsamuel Do you want to modify this PR or close it?

girardinsamuel:
I will edit this PR and open a PR on ORM repo too.

girardinsamuel:
First we should merge this into ORM repo: https://github.com/MasoniteFramework/orm/pull/629
Then, tests should pass in this PR.

josephmancuso (approved):

PR #517, opened 2022-01-25
Add missing AuthenticationMiddleware

girardinsamuel:
from M3.


josephmancuso (commented):
i dont know ... one of the reasons its in the cookie cutter repo is because the redirect is customizable.

girardinsamuel:
@josephmancuso maybe adding it in the project template is better then ?
Here it is: https://github.com/MasoniteFramework/cookie-cutter/pull/37

josephmancuso:
Yeah I think putting it in cookie cutter was a better idea

PR #523, opened 2022-02-06
Add `ip()` helper to request

girardinsamuel:
Allow to retrieve client IP easily from request instance.


josephmancuso (changes requested):
Getting the IP Address is tricky I think we need to take into consideration `HTTP_X_FORWARDED_FOR ` https://stackoverflow.com/questions/7835030/obtaining-client-ip-address-from-a-wsgi-app-using-eventlet

girardinsamuel:
> Getting the IP Address is tricky I think we need to take into consideration `HTTP_X_FORWARDED_FOR ` https://stackoverflow.com/questions/7835030/obtaining-client-ip-address-from-a-wsgi-app-using-eventlet

Hmm you're right !
Laravel is using Symfony function to handle it and here is how Symfony is handling it:
https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/HttpFoundation/Request.php#L758

Interesting ! They indeed use `HTTP_X_FORWARDED_FOR`

josephmancuso:
It might be good to maybe somehow take into account a user defined order of getting the IP? Thinking in the event that some Azure or AWS setup has custom proxy or something.

Maybe make this a middleware?

See this answer:

```php
public function getIp(){
    foreach (array('HTTP_CLIENT_IP', 'HTTP_X_FORWARDED_FOR', 'HTTP_X_FORWARDED', 'HTTP_X_CLUSTER_CLIENT_IP', 'HTTP_FORWARDED_FOR', 'HTTP_FORWARDED', 'REMOTE_ADDR') as $key){
        if (array_key_exists($key, $_SERVER) === true){
            foreach (explode(',', $_SERVER[$key]) as $ip){
                $ip = trim($ip); // just to be safe
                if (filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE) !== false){
                    return $ip;
                }
            }
        }
    }
    return request()->ip(); // it will return server ip when no client ip found
}
```

https://stackoverflow.com/questions/33268683/how-to-get-client-ip-address-in-laravel-5

josephmancuso:
Maybe we make it a middleware like this:

```python
class IPMiddleware(Middleware):

    ___headers__ = [
         'HTTP_CLIENT_IP', 
         'HTTP_X_FORWARDED_FOR', 
         'HTTP_X_FORWARDED', 
         'HTTP_X_CLUSTER_CLIENT_IP', 
         'HTTP_FORWARDED_FOR', 
         'HTTP_FORWARDED', 
         'REMOTE_ADDR'
    ]

    #..
```

and you can inherit the middleware and overwrite this list if you needed a custom order lookup?

girardinsamuel:
> Maybe we make it a middleware like this:
> 
> ```python
> class IPMiddleware(Middleware):
> 
>     ___headers__ = [
>          'HTTP_CLIENT_IP', 
>          'HTTP_X_FORWARDED_FOR', 
>          'HTTP_X_FORWARDED', 
>          'HTTP_X_CLUSTER_CLIENT_IP', 
>          'HTTP_FORWARDED_FOR', 
>          'HTTP_FORWARDED', 
>          'REMOTE_ADDR'
>     ]
> 
>     #..
> ```
> 
> and you can inherit the middleware and overwrite this list if you needed a custom order lookup?

That's smart I like it ! So the middleware would add the ip in the request ? How would you add `request.ip()` from the middleware ?

josephmancuso:
umm. we can add an `ip` method to the request that returns `request._ip` and the middleware would just populate the `_ip` attribute

girardinsamuel:
What do you think of this ?

josephmancuso:
Looks good. Ill take a better look tonight

josephmancuso (approved):

josephmancuso (approved):

PR #525, opened 2022-02-08
Add `application` to Jinja2 environment to make it accessible in custom extensions

girardinsamuel:
In Jinja2 extensions that we can develop we can now do:
```python

class MyExtension(Extension):
# ...
    self.environment.application  # == Masonite app container
```

### DIFF
--- a/src/masonite/views/view.py
+++ b/src/masonite/views/view.py
@@ -255,6 +255,9 @@ class View:
             extensions=self._jinja_extensions,
             line_statement_prefix="@",
         )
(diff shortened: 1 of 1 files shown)


josephmancuso (approved):

josephmancuso:
How do Jinja 2 extensions for Masonite work?

girardinsamuel:
> How do Jinja 2 extensions for Masonite work?

I built one for a new package I will show you once it's done 😉

PR #527, opened 2022-02-10
Add Rate Limiting feature

girardinsamuel:
This PR adds Rate Limiting feature to Masonite 🚀:
- It can be used for requests with the `ThrottleRequestsMiddleware` (e.g. to throttle an api endpoint)
- It can be used for jobs or any other cases where you need to rate limit usage, you can even use it to rate limit a simple python callable.

Documentation can be found here https://github.com/MasoniteFramework/docs/pull/183


tests/features/event/test_event.py
```diff
@@ -56,13 +56,15 @@ def setUp(self):
         self.event = self.application.make("event")
         self.event.listen(UserAddedEvent, [SendEmailListener])
 
+    def tearDown(self):
```
girardinsamuel, line 59: Here I am resetting events that are listened to have idempotent tests and to avoid having those events in other unit tests.

src/masonite/cache/drivers/FileDriver.py
```diff
@@ -25,7 +25,7 @@ def add(self, key, value, seconds=None):
 
     def get(self, key, default=None, **options):
         if not self.has(key):
-            return None
+            return default
```
girardinsamuel, line 28: Just a small fix for getting cache key

src/masonite/facades/Cache.py
```diff
@@ -0,0 +1,5 @@
+from .Facade import Facade
```
girardinsamuel, line 1: add missing cache facade

src/masonite/facades/RateLimiter.pyi
```diff
@@ -0,0 +1,43 @@
+from typing import Any, Callable, TYPE_CHECKING
```
girardinsamuel, line 1: Here I discovered a way to type hint facades so that e.g. in VS Code when using facade we get completion as we were using the class directly !! It's so nice !
So in the type hinting PR, I will add those type hints for all the facades 😉

josephmancuso: awesome!!!!!!! that was a huge pain for me

josephmancuso: Do we have to create a pyi file for each facade?

josephmancuso (commented):
Started to review this. Got halfway. Will continue later

src/masonite/facades/RateLimiter.pyi
girardinsamuel: Yes..that's the counterpart. But not a big deal IMO. We are not adding facades every day, and we will just need to take care during code review of new features, that a facade (if necessary) exists and that its `pyi` counterpart exists too.

It's a documentation on its own so it's as important as maintaining doc on the site IMO.

josephmancuso:
Fixed merge conflicts

PR #529, opened 2022-02-14
Changing signature of Gate.register_policies

Marlysson:
Fix #529 
Changing the signature of 'Gate.register_policies' to receive a Policy class with model related injected in constructor.

### DIFF
--- a/src/masonite/authorization/Gate.py
+++ b/src/masonite/authorization/Gate.py
@@ -28,9 +28,9 @@ class Gate:
 
         self.permissions.update({permission: condition})
 
-    def register_policies(self, policies):
-        for model_class, policy_class in policies:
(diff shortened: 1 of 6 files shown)


josephmancuso:
@Marlysson @girardinsamuel This is open for discussion. This came out of a discord conversation. I thought it was weird registering policies by using a list of tuples rather than a more elegant syntax

girardinsamuel:
> @Marlysson @girardinsamuel This is open for discussion. This came out of a discord conversation. I thought it was weird registering policies by using a list of tuples rather than a more elegant syntax
I agree, I will take a look closer as it's something I have implemented.

girardinsamuel:
As it's a change of signature and potentially a breaking change, we will move this to Masonite 5.

PR #533, opened 2022-02-15
Fix notification mail driver to use default driver from `config/mail.py`

### LINKED ISSUE #532: Mail notification driver is not using default mail driver defined in `config/mail.py`
I am trying to implement notification feature in my application but it looks like the feature is yet in developing phase . 
Procedure:
1. Firstly, I created a notification Test
2. Then, used the channel mail for notification
3. When I used to_mail method inside notification, it printed mail in terminal.

Could  you please  guide  me how it can be resolved . 

When I went through the code, I found following code in mail driver:
Screenshot from 2022-02-15 11-10-33

yubarajshrestha:
Fixes notification issue with mail driver.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,3 @@ src/masonite.egg-info/*
 .vscode
 build/
 venv4
-config/
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ src/masonite.egg-info/*
 .vscode
 build/
 venv4
+config/
--- a/src/masonite/mail/MockMail.py
+++ b/src/masonite/mail/MockMail.py
@@ -1,12 +1,16 @@
 from .Mail import Mail
+from ..configuration import config
 
 
 class MockMail(Mail):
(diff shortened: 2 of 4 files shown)


girardinsamuel:
We don't have any options to define mail driver to use for notification mail driver so we cannot set that in config and we do not want to have duplicates places where we set a mail driver...

However, I agree it would be great 👍 to be able to override the mail driver used to send mail notification on the fly.
For now it's possible when definining the mail notification because we are using a `Mailable`:

```python
class CustomNotification(Notification, Mailable):
    def to_mail(self, notifiable):
        return (
            self.subject("Masonite 4")
            .from_("joe@masoniteproject.com")
            .text("Hello from Masonite!")
            # Here I am overriding the driver with `mailgun` else it will use default mail driver from `config/mail.py`.
            .driver("mailgun")
        )

    def via(self, notifiable):
        return ["mail"]
```

So I edited your PR to simply use default mail driver defined in config/mail.py and updated the tests to make it cleaner.
Fixes #532

PR #540, opened 2022-02-23
Fix creating events, listeners and tasks on Linux

### LINKED ISSUE #539: Unable to create task
### Describe the bug

I am not able to create task using `craft` command.

### Expected behaviour

When I run `python craft task TaskName` a TaskName file should be created. But it's not.

1. Create masonite project
2. Try to create task using craft command.

### Screenshots

```sh
FileNotFoundError

  [Errno 2] No such file or directory: '/home/kodiary/python_project/micro_service/nt/venv/lib/python3.9/site-packages/masonite/scheduling/commands/../../stubs/scheduling/Task.py'

  at venv/lib/python3.9/site-packages/masonite/utils/filesystem.py:65 in render_stub_file
      61│ 
      62│ 
      63│ def render_stub_file(stub_file, name):
      64│     """Read stub file, replace placeholders and return content."""
    → 65│     with open(stub_file, "r") as f:
      66│         content = f.read()
      67│         content = content.replace("__class__", name)
      68│     return content
      69│ 
```

### OS

Linux

### OS version

Ubuntu 18.04

### Browser

_No response_

### Masonite Version

4.6.3

### Anything else ?

:(

girardinsamuel:
On Linux path manipulation is case sensitive and stubs file were defined with incorrect case.
Fix #539


josephmancuso (approved):

josephmancuso:
had a feeling a file path was the issue. Linux is very strict on file path case sensitivity. We have same issue as work between Macs and Linux

josephmancuso:
@girardinsamuel if you can release this into a patch fix if you have time

PR #547, opened 2022-02-27
added list defaults when using [] in inputs

### LINKED ISSUE #544: Return empty list when using file input fields as list and no file is selected
### Describe the bug

Return an empty list when using `<input type="file" name="some_file[]">` and no file is selected.

### Expected behaviour

Return empty list when no file selected

_No response_

### Screenshots

_No response_

### OS

Linux

### OS version

KDE 5.24.1

### Browser

_No response_

### Masonite Version

4.7.0

### Anything else ?

Possible fix could be

_InputBag.py_
```
elif isinstance(value, bytes):
    file = []
    if fields[name].filename:
        file = [UploadedFile(fields[name].filename, value)]
    self.post_data.update(
        {name: file}
    )
```

josephmancuso:
Closes #544

### DIFF
--- a/src/masonite/input/InputBag.py
+++ b/src/masonite/input/InputBag.py
@@ -99,6 +99,9 @@ class InputBag:
                     )
 
     def get(self, name, default=None, clean=True, quote=True):
+        if isinstance(name, str) and name.endswith("[]"):
+            default = []
+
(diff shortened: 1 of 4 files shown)


josephmancuso (commented):
approved

PR #551, opened 2022-03-04
Fix `actingAs` test helper during requests

### LINKED ISSUE #550: `actingAs` tests helper does not log user in when using get,post... test methods
### Describe the bug

when doing in a test
```python
self.actingAs(User.first()).get("/")
```

user is not logged in during the request to "/".

**What is happening ?**
I think it's because when using `get("/")` in a test the request will be overriden and then the user logged in won't be logged in anymore.

### Expected behaviour

The user should be logged in.

_No response_

### Screenshots

_No response_

### OS

macOS

### OS version

Monterey 12.3

### Browser

_No response_

### Masonite Version

4.7.0

### Anything else ?

_No response_

girardinsamuel:
Fix #550
- [x] Fix `actingAs` method
- [x] Add unit test to show the problem
- [x] Fix the problem

⚠️ Last issue to address is the following (which is why one test is failing):

Imagine the scenario where we want to test a login view where we try to log the user in using the web guard. We would like to assert that user is authenticated when the request completes.
```python
# the controller method for the url /test-authenticates

def auth(self, request: Request):
        request.app.make("auth").guard("web").attempt("idmann509@gmail.com", "secret")
        return "logged in"
```
And the corresponding test:
```python
self.get("/test-authenticates").assertAuthenticated()
```

This test will fail because the `assertAuthenticated` assertion will use web guard to check if the user is logged in. For this it will check if response contains a `token` cookie which value matches a user `remember_token` in users table.

👉 In real life this is working, but in tests when we receive back the response cookies are encrypted and the comparison of the tokens will fail because we will check remember_token against its encrypted value.

Of course we could use `assertRequestAuthenticated()` which is working because it uses user loaded into the request to assert that user is authenticated, but it would be nice that the first method worked too !


girardinsamuel:
For now I am still having issue, it's not an easy task...I have to look at it again

girardinsamuel:
@josephmancuso if you have some suggestions or ideas about the above problem it would be really nice !

src/masonite/tests/TestCase.py
```diff
@@ -63,6 +65,9 @@ def setUp(self):
 
         self.routes_to_restore = set(self.application.make("router").routes)
 
+        # don't encrypt cookies during tests
```
girardinsamuel: First solution would be to disable cookies encryption during tests

girardinsamuel: But then CSRF token verif will fail... so that's not a good solution IMO.

PR #553, opened 2022-03-05
Added feature to share individual input validation errors in session

### LINKED ISSUE #552: Validation Error Message for individual input fields.

There's no any way to get the validation error message for the individual form fields.

We have option get list all the validation error message if exists.
```jinja
@if session().has('errors'):
    <div class="bg-yellow-400">
        <div class="bg-yellow-200 text-yellow-800 px-4 py-2">
            <ul>
                @for key, error_list in session().get('errors').items():
                @for error in error_list
                <li>{{ error }}</li>
                @endfor
                @endfor
            </ul>
        </div>
    </div>
@endif
```

The solution is to get above response as well as ability to get individual error message like below:
```jinja
<div>
                <label for="name">Name</label>
                <input type="text" name="name" placeholder="Name">
                @if errors.has('name')
                <code>{{ errors.get('name')[0] }}</code>
                @endif
</div>
```
and the different approach of getting all errors:
```jinja
 @if errors.any()
    <div class="alert alert-danger">
        <ul>
            @for key, message in errors.all().items()
            <li>{{ message }}</li>
            @endfor
        </ul>
    </div>
@endif
```

_No response_

- [ ] Yes

### Anything else ?

_No response_

yubarajshrestha:
Fixes #552.


girardinsamuel (changes requested):
IMO, this would be better to do it in its own middleware to allow a user to disable sharing validation errors.

Also errors are already shared in session through `errors` key. I would directly use this key instead of using `validation_errors` key.

So maybe, we should mix both of those approach in `ShareErrorsInSession` middleware:

- Put sharing errors logic in this middleware
- Map errors with MessageBag as you did.

We would end up with something like this:
```python

class ShareErrorsInSessionMiddleware(Middleware):
    def before(self, request, response):
         request.app.make("view").share(
                {
                    "errors": MessageBag(Session.pull("errors") or {}),
                }
            )
        return request

    def after(self, request, _):
        return request
```

yubarajshrestha (commented):
Created separate middleware to handle errors in frontend...

src/masonite/middleware/route/ShareErrorsInSessionMiddleware.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class ShareErrorsInSessionMiddleware(Middleware):
+    """This part verifies if there are any errors stored in Session.
+    If there is a validation error then it shares the "errors" key to display the errors in frontend
+    else it will return an empty MessageBag object.
+    """
+
+    def before(self, request, _):
+        request.app.make("request").session = Session
```
girardinsamuel: You should not need to do this 🤔

src/masonite/views/view.py
```diff
@@ -279,3 +280,9 @@ def get_response(self):
     def test(self, key, obj):
         self._tests.update({key: obj})
         return self
+
+    def reset_vaidation_errors(self):
```
girardinsamuel:
```suggestion
    def reset_validation_errors(self):
```

girardinsamuel (changes requested):

src/masonite/middleware/route/ShareErrorsInSessionMiddleware.py
yubarajshrestha: Ah Fixed...

src/masonite/views/view.py
yubarajshrestha: Actually we do not need that part anymore so I am removing that method.

yubarajshrestha (commented):
Fixed requested changes.

girardinsamuel:
I took a look again here and I would like to add a unit test.
So I added a unit test and I got an error..I need to look a bit into it, I will try to solve this soon !

josephmancuso:
whats going on with this PR? is it good?

girardinsamuel:
Hmm for now there is something with the unit test I wrote, I don't know if the issue is coming from tests or from the framework itself. I will push this so we can discuss it.

yubarajshrestha:
@josephmancuso @girardinsamuel what's the status for this PR? Any issues? It's working perfectly in my case in all of my applications.

girardinsamuel:
> what's the status for this PR? Any issues? It's working perfectly in my case in all of my applications.

I was waiting for the last release to be made, to hopefully fix the unit tests I pushed to this PR.
And I will need to review it again.

josephmancuso:
Put this forked branch up to date with 4.0

josephmancuso:
> @josephmancuso @girardinsamuel what's the status for this PR? Any issues? It's working perfectly in my case in all of my applications.

@yubarajshrestha what's preventing me from merging it really is that failing test.

girardinsamuel:
It's a test I have added. I will remove it for now

girardinsamuel (approved):
That's okay for me for now. 
In Masonite 5, maybe we will need to refactor session flashing and message bag validation features.

josephmancuso (approved):

PR #559, opened 2022-03-10
Fix parsing post data

girardinsamuel:
- [x] add failing test
- [x] fix behaviour

I have added a test for each content type including the one which was not working (i.e. `multipart/form-data`) so that we can better prevent regression errors.
It would maybe need more tests as it's dealing with user inputs and there are lot of edge cases, but it should be good for now.

### DIFF
--- a/src/masonite/input/InputBag.py
+++ b/src/masonite/input/InputBag.py
@@ -56,11 +56,6 @@ class InputBag:
                 self.post_data = self.parse_dict(parsed_request_body)
 
             elif "multipart/form-data" in environ.get("CONTENT_TYPE", ""):
-                try:
-                    request_body_size = int(environ.get("CONTENT_LENGTH", 0))
-                except (ValueError):
-                    request_body_size = 0
-
                 fields = cgi.FieldStorage(
                     fp=environ["wsgi.input"],
                     environ=environ,
--- a/src/masonite/input/InputBag.py
+++ b/src/masonite/input/InputBag.py
@@ -201,9 +201,9 @@ class InputBag:
                     else:
                         d.setdefault(gd["name"], {})[gd["value"]] = value[0]
                 else:
-                    try:
+                    if isinstance(value, (list, tuple)):
                         d.update({name: value[0]})
-                    except TypeError:
+                    else:
                         d.update({name: value})
 
         new_dict = {}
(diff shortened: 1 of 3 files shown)


src/masonite/input/InputBag.py
```diff
@@ -56,11 +56,6 @@ def parse(self, environ):
                 self.post_data = self.parse_dict(parsed_request_body)
 
             elif "multipart/form-data" in environ.get("CONTENT_TYPE", ""):
-                try:
```
girardinsamuel, line 59: not used

josephmancuso: hmm. How do you know this isn't used? We may just not have a test for it.

I would assume in the event the browser does not pass a content length for the request this would throw an error right?

tests/core/request/test_input.py
```diff
@@ -95,6 +126,21 @@ def test_can_parse_post_params_from_url_encoded(self):
...
 
+    def test_can_parse_multipart_formdata_content_type(self):
+        data, content_type = encode_multipart_formdata({"key": "value", "test": 1})
+        bag = InputBag()
+        bag.load(
+            {
+                "REQUEST_METHOD": "POST",
+                "CONTENT_TYPE": content_type,
+                "CONTENT_LENGTH": str(len(data.encode("latin-1"))),
+                "wsgi.input": io.BytesIO(data.encode("latin-1")),
```
josephmancuso: why latin-1 here and not utf-8?

josephmancuso (changes requested):

src/masonite/input/InputBag.py
girardinsamuel: I mean that the variable in the code is really not used later.
That's the difference with the other content types where we are using it.

tests/core/request/test_input.py
girardinsamuel: It's just that I used an example found on internet to simulate a multipart/form-data (took me hours to find this).
But maybe it will work with utf-8, I will check.

src/masonite/input/InputBag.py
josephmancuso: Oh ok cool lol

tests/core/request/test_input.py
girardinsamuel: that's fixed

josephmancuso (approved):

PR #561, opened 2022-03-10
Fix merging Mailable options with general `mail` options

### LINKED ISSUE #558: Incorrect .from_('') field when sending emails
### Describe the bug

When sending emails, I'm getting the wrong email address in the from field. in the build I have,  .from_(noreply@aspent.org) and this is the email that I want be in the from field

Screen Shot 2022-03-09 at 12 39 23 AM
.

However the email i'm getting in the from field when the email is sent, is a different one, which looks like is coming from the smtp configuration.

Screen Shot 2022-03-09 at 12 44 17 AM

### Expected behaviour

The .from_('') field in the build file should be taking precedence, not the one in the smtp configuration file.

1. Create a Mailable 
2. Provide the .from_('') with an email you want to show in the from field.
3. send the email to test.

### Screenshots

_No response_

### OS

macOS

### OS version

Monterey 12.1

### Browser

_No response_

### Masonite Version

4.6.3

### Anything else ?

_No response_

girardinsamuel:
Fix #558.

The issue is that mailable `from` option is always overriding from options defined for the selected driver in `config/mail.py`. 
So when no from options is given in the mailable, the value of from is "" and the default option from the selected driver cannot be used.

In addition to this, MockMail was not provided with the mail configuration at initialization and the mail config of the test project was not up to date. This has been updated accordingly.

Tests have been improved a bit to do more real testing, e.g. SMTP and Mailgun drivers are really tested (still as an integration test) and their error handled. At least we can really test sending the email even if no credentials were provided.

### DIFF
--- a/tests/features/mail/test_mailable.py
+++ b/tests/features/mail/test_mailable.py
@@ -70,17 +70,6 @@ class TestMailable(TestCase):
             "Joseph Mancuso <idmann509@gmail.com>, <joe@masoniteproject.com>",
         )
 
-    def test_recipient(self):
-        to = Recipient("idmann509@gmail.com, joe@masoniteproject.com")
-        self.assertEqual(
-            to.header(), "<idmann509@gmail.com>, <joe@masoniteproject.com>"
-        )
-        to = Recipient("Joseph Mancuso <idmann509@gmail.com>, joe@masoniteproject.com")
-        self.assertEqual(
-            to.header(),
-            "Joseph Mancuso <idmann509@gmail.com>, <joe@masoniteproject.com>",
-        )
-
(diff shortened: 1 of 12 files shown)


tests/features/mail/test_mailable.py
```diff
@@ -59,17 +59,6 @@ def test_attach(self):
             == 1
         )
 
-    def test_recipient(self):
-        to = Recipient("idmann509@gmail.com, joe@masoniteproject.com")
```
girardinsamuel, line 63: this is present in the test below this one.

josephmancuso (approved):

PR #563, opened 2022-03-13
Fix deprecation warning for `distutils` module in craft `auth` command

### LINKED ISSUE #562: Fix deprecation warning with `distutils` module
`distutils` module will be deprecated in Python `3.12`.
```
src/masonite/commands/AuthCommand.py:2
  /home/runner/work/masonite/masonite/src/masonite/commands/AuthCommand.py:2: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
    from distutils.dir_util import copy_tree
```
For now we are using it in `AuthCommand`: 
```python
from distutils.dir_util import copy_tree
```
We should be able to use `shutil.copytree` instead.

girardinsamuel:
Fix #562.

Old `copy_tree` method doc:
https://docs.python.org/3/distutils/apiref.html#distutils.dir_util.copy_tree

New `copytree` method doc:
https://docs.python.org/3/library/shutil.html#shutil.copytree

The new method will complain if directory already exists when the old method was not complaining. To keep same behaviour we need to add, `dirs_exists_ok=True` argument to allow overriding. But this option is only available starting from Python 3.8. As Masonite is compatible with Python 3.7, we need to handle this case (this switch will be removed when 3.7 will be dropped by Masonite one day).

I am not a big fan of switches but here it's not that much of a problem IMO.

### DIFF
--- a/src/masonite/commands/AuthCommand.py
+++ b/src/masonite/commands/AuthCommand.py
@@ -1,5 +1,5 @@
 """Scaffold Auth Command."""
-from distutils.dir_util import copy_tree
+from shutil import copytree
 import os
 
 from ..utils.location import controllers_path, views_path, mailables_path
@@ -19,13 +19,13 @@ class AuthCommand(Command):
         self.app = application
 
     def handle(self):
-        copy_tree(
+        copytree(
             self.get_template_path(),
             views_path("auth"),
         )
-        copy_tree(self.get_controllers_path(), controllers_path("auth"))
+        copytree(self.get_controllers_path(), controllers_path("auth"))
 
-        copy_tree(self.mailables_path(), mailables_path())
+        copytree(self.mailables_path(), mailables_path())
 
         self.info("Auth scaffolded successfully!")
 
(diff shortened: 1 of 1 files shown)


src/masonite/commands/AuthCommand.py
```diff
@@ -19,13 +20,21 @@ def __init__(self, application):
...
+                self.get_controllers_path(),
+                controllers_path("auth"),
+                dirs_exist_ok=True,
+            )
+
+            copytree(self.mailables_path(), mailables_path(), dirs_exist_ok=True)
+        else:
+            copytree(self.get_template_path(), views_path("auth"))
+            copytree(self.get_controllers_path(), controllers_path("auth"))
+            copytree(self.mailables_path(), mailables_path())
```
josephmancuso, line 37: Why don't we need `dirs_exist_ok` here?

josephmancuso (changes requested):

src/masonite/commands/AuthCommand.py
girardinsamuel: Because the feature (and the argument) does not exist in Python < 3.8.
So for Python 3.7, we will loose the behaviour of automatically overriding the folders if they exist already !

josephmancuso: ooh ok I see whats going on here

PR #570, opened 2022-03-24
fixed environment overriding issue

josephmancuso:
Fixes a few issues I found when testing

### DIFF
--- a/src/masonite/environment/environment.py
+++ b/src/masonite/environment/environment.py
@@ -32,7 +32,7 @@ class LoadEnvironment:
         if environment:
             self._load_environment(environment, override=override)
 
-        if "pytest" in sys.modules:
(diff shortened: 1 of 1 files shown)


src/masonite/environment/environment.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, environment=None, override=True, only=None):
         if environment:
             self._load_environment(environment, override=override)
 
-        if "pytest" in sys.modules:
```
josephmancuso, line 35: This code is 3 years old lol. The issue is that if pytest is imported anywhere before this load environment class is called, then this will return true (which we do I think inside Masonite code). So this was loading the testing env during normal development

girardinsamuel (approved):

PR #577, opened 2022-04-07
Added more features to `CORS`

### LINKED ISSUE #476: Cors header should support multiple origins
One of the issues with CORS multiple origins is that only 1 `Access-Control-Allow-Origin` value is allowed. So you can't just do 

```
"Access-Control-Allow-Origin": "server1.com,server2.com"
```

You need to either return server1 or server2. You can can the value in an environment variable on the server or something but usually what you want to do is:

```
"hosts": ["server1.com","server2.com"]
```

So we will need some options to set. We can do this in a backwards compatible way.

It might be worth it to look at options in other frameworks:

```
'paths' => ['api/*', 'sanctum/csrf-cookie'],
'allowed_methods' => ['*'],
'allowed_origins' => ['*'],
'allowed_origins_patterns' => [],
'allowed_headers' => ['*'],
'exposed_headers' => [],
'max_age' => 0,
'supports_credentials' => false,
```

Once these options are set you we can update the options to these in the framework under the `self.options` attribute<br />[#tara-label-#tara-label-Tara Task]<br />[Tara Task]<br />[Tara Task](https://app.tara.ai/masonite-x-inc/*/tasks/43)

girardinsamuel:
This PR improve Masonite CORS operations protection capabilities (Fix #476)

- [x] Add a real `cors` feature as any other Masonite feature
- [x] Use this feature in the existing `CorsMiddleware`
- [x] Add unit tests
- [x] Add docs: you can find a preview here https://docs.masoniteproject.com/v/development/security/cors
- [x] Add config to new projects: https://github.com/MasoniteFramework/cookie-cutter/pull/46


josephmancuso:
Agreed. I like where this is heading

src/masonite/routes/HTTPRoute.py
```diff
@@ -30,7 +29,12 @@ def __init__(
         self.controller_method = None
         self._domain = None
         self._name = name
+
         self.request_method = [x.lower() for x in request_method]
+        # add HEAD method if GET
+        if "get" in self.request_method and "head" not in self.request_method:
```
josephmancuso: This part is confusing to me. Is there another way to do this? Seems weird automatically adding head to every GET route

src/masonite/routes/Route.py
```diff
@@ -24,7 +24,7 @@ def get(self, url, controller, module_location=None, **options):
         return HTTPRoute(
             url,
             controller,
-            request_method=["get"],
+            request_method=["get", "head"],
```
josephmancuso: Not sure i like doing this. This feature seems to tightly coupled to other features

tests/features/cors/test_cors.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+class TestCors(TestCase):
+    def setUp(self):
+        super().setUp()
+        # add CorsMiddleware
+        self.application.make("middleware").add([CorsMiddleware])
+        self.cors = self.application.make("cors")
+        self.default_options = {
+            "allowed_origins": ["http://localhost"],
```
josephmancuso, line 13: Good to whichever origin matches an element in this list, thats the host that is returned in the origin header right?

tests/integrations/config/cors.py
```diff
@@ -0,0 +1,15 @@
+"""Cross-Origin Resource Sharing (CORS) config"""
+
+PATHS = ["api/*", "sanctum/csrf-cookie", "server1.com"]
+
+ALLOWED_METHODS = ["*"]
+
+ALLOWED_ORIGINS = ["*"]
```
josephmancuso: I think we need to have these configs more similar to our other configs. 

Something like:

```
CORS = {
   "paths": ["api/*", "sanctum/csrf-cookie", "server1.com"],
   "allowed_methods": ["*"],
   ....
   
}
```

josephmancuso (changes requested):

src/masonite/routes/HTTPRoute.py
girardinsamuel: That 's how Laravel is handling it and that's quite mandatory because each GET route should be able to be accessed through GET !

tests/features/cors/test_cors.py
girardinsamuel: I did not understand the question

tests/integrations/config/cors.py
girardinsamuel: The issue is with this is that it's already namespaced with cors through the config file name...so accessing it later will result in doing `config("cors.cors.paths")`....

src/masonite/routes/Route.py
girardinsamuel: Above item and this one should be discussed in the PR https://github.com/MasoniteFramework/masonite/pull/578.

The HEAD thing is a prerequisite to this PR and is implemented in #578.

josephmancuso:
@girardinsamuel if you can, can you resolve all these conflicts? I merged in the PR for the options and HEAD

girardinsamuel:
It's done. I just need to update documentation.
And maybe the way cors options are defined as you said in your review.

tests/features/cors/test_cors.py
josephmancuso: I have no idea what I was trying to ask here lol

tests/integrations/config/cors.py
josephmancuso: The rest of the PR looks good but still not sold on this.

> The issue is with this is that it's already namespaced with cors through the config file name...so accessing it later will result in doing `config("cors.cors.paths")`....

You have a valid point here. Honestly it looks better than this config file though. This looks very M3 to me.

Just throwing some idea out there but do we have like a `security.py` file so at least it would be `config("security.cors.paths")`? And we can put other security settings in there either now or later

josephmancuso:
Just not sold on the config file

tests/integrations/config/cors.py
girardinsamuel: Hmm we don't have it yet. But we could add it, `cors` was already new so we could change it to `security` why not, i like it

girardinsamuel:
> Just not sold on the config file

I updated to what you proposed. cookie-cutter and docs are updated too
👉 https://docs.masoniteproject.com/v/development/security/cors#configuration you're right it's cleaner

josephmancuso (approved):
Code review looks good. Will need to test this manually.

One thing I want to be sure of is what happens when there is no security.py file

girardinsamuel:
> Code review looks good. Will need to test this manually.
> 
> One thing I want to be sure of is what happens when there is no security.py file

You're right. I have added some default options for old projects ✅. In Masonite 5 we will be able to remove the lines I added. And maybe create a SecurityProvider for registering `cors` feature.

josephmancuso (approved):

PR #578, opened 2022-04-08
Add ability to handle correctly `OPTIONS` and `HEAD` requests

girardinsamuel:
This PR adds:

## OPTIONS request handling
Before, if a route with `OPTIONS` had not been defined, a request to an existing route with OPTIONS method was returning a 404 Not Found. Now, it's returning a 204 No Content response with `Allow` headers containing the allowed request methods for this route.

(ref https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS#identifying_allowed_request_methods)

## Method Not Allowed handling

Before if a route was accessed with wrong request method, it was returning a 404 Not Found. Now if the route exists but accessed with wrong methods, it will return a 405  Method Not Allowed exception with error message containing the allowed request methods for this route.

## HEAD request handling
Before, if a route with `HEAD` had not been defined, a request to an existing route with HEAD method was returning a 404 Not Found. Now, it's returning a 204 No Content response with the headers that would have been sent if the request had been made with `GET` method.

(ref https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD)

In addition, when defining a `GET` route, Masonite will automatically add a `HEAD` route for it.

Here is a summary with CURL commands against a Masonite app:
image

For this, two small features have been added:
- adds `withHeaders` to response to add headers as a dict easily, as `header()` was one header by one and was not returning response class
- adds ability to add headers to HTTP exception when rendering exceptions.


src/masonite/routes/HTTPRoute.py
```diff
@@ -191,6 +197,15 @@ def get_response(self, app=None):
             params = self.extract_parameters(app.make("request").get_path()).values()
             # Resolve Controller Method
             response = app.resolve(getattr(controller, self.controller_method), *params)
+
+            # if request method is HEAD, return a response without content but with
+            # headers that would be returned if request's URL was requested with GET method instead
+            # https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD
+            if app.make("request").get_request_method().upper() == "HEAD":
```
girardinsamuel, line 228: Here we should get the response headers and build a new response with the same headers but without content and the 204 status code. Also the response bound to the container is not good anymore we have to replace it.

and here I am not really satisfied on the way to override the current response and build a new one... how could that be written better ?

src/masonite/routes/Route.py
```diff
@@ -24,7 +24,7 @@ def get(self, url, controller, module_location=None, **options):
         return HTTPRoute(
             url,
             controller,
-            request_method=["get"],
+            request_method=["get", "head"],
```
josephmancuso, line 28: what is this for?

josephmancuso (changes requested):

src/masonite/routes/Route.py
girardinsamuel: A framework should respond to all HEAD requests on its defined routes. So whenever you define a route with GET it will actually register a route with HEAD and a route with GET.

girardinsamuel: It's the expected behaviour for web frameworks : https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD

girardinsamuel:
@josephmancuso can you take a look at this? This PR is required for the CORS feature 😉.

josephmancuso (approved):

josephmancuso:
code review good. Just want to test this manually before merging

PR #593, opened 2022-04-16
Adding check for logged in users in register, login Controllers

theokafadaris:
Checking if user is logged in for Login, Register. If yes, then user is redirected to /home.


girardinsamuel:
@josephmancuso shouldnt we use a GuestsOnlyMiddleware or something like this instead ?

josephmancuso:
Hmm I like the idea of a middleware for this

josephmancuso:
We could make a middleware and then call it like guests:route_name for the redirection

The only issue is that's not in people projects right now so it wouldn't work doing it that way.

Maybe we just hold off on this and then do it the right way in V5?

girardinsamuel:
Yes I a guess we have to propose a more featured (with necessary requirements that people are expecting) auth in v5

josephmancuso (approved):

josephmancuso:
gonna close this PR. Thinking more about this, the features added to this PR can easily be done in someones application and the correct solution here should be done as middleware in Masonite V5

PR #597, opened 2022-04-19
Fixed typo

dwedigital:
Changed to responsible in the error message returned


dwedigital:
Turns out I was wrong and responsable is a made up word referring to the method handling the response!

PR #605, opened 2022-04-25
Add possibility to add method to controller instances

girardinsamuel:
In latest controller binding changes, we forgot to add the possibility to use a method for controller instances instead of the default `__call__` binding. Now we can do:

```python
class SomeController(Controller):

    def show(self):
         # ...

controller = SomeController()

Route.get("/", controller.show)
```


girardinsamuel (commented):
Reviewed

PR #607, opened 2022-04-28
Switch to `wsgiref` standard python development server

### LINKED ISSUE #606: Configure dev server to avoid various issues in development
Some users reported different glitches related to serving the app with `waitress`.

We should either better configure the dev server or upgrade it to something else.

We could have a look into what's Django is using...

girardinsamuel:
Trying to fix #606


josephmancuso:
The other implementation looks cleaner 👍

PR #616, opened 2022-05-05
Feature/613 - Added Route Views

### LINKED ISSUE #613: Add Route Views

Route views are a way where you can specify a view directly instead of needing to create a controller method for it

Need to create a controller method and return a view

Need to have something like this:

```python
ROUTES = [
  Route.view('/route', 'view.name', {"key": "value"})
]
```

_No response_

- [ ] Yes

### Anything else ?

_No response_

josephmancuso:
Closes #613


josephmancuso (commented):
reviewed

josephmancuso (commented):
reviewed

PR #620, opened 2022-05-07
Allow unbound class resolution from container

### LINKED ISSUE #618: Allow unbound class resolution from container

Currently, resolving a dependency from the container requires previously binding that class in some manner explicitly, even if all that is required to "resolve" the class is to simply instantiate it as `AnyClass()`.

Any sufficiently large and complex application will require dozens of such rote bindings.

**The proposal here is to make container resolution have a simple form of non-bound resolution as the fallback when an explicit binding does not exist.**

This would improve developer experience by removing the code needed to add a `.simple()` binding somewhere, likely inside a ServiceProvider.

Moreover, this already partially exists in Controller resolution within HTTPRoute:

```python
self.controller_class = Loader.get_object(
    controller.__module__, controller_name, raise_exception=True
)

...

controller = self.controller_class()

return getattr(controller, self.controller_method)()
```

Exception case:

```python
class AnyClass:
    pass

# never bound via self.application.bind/simple/etc
instance = self.application.make(AnyClass)

# src.masonite.exceptions.exceptions.MissingContainerBindingNotFound: The dependency with the <class 'AnyClass'> annotation could not be resolved by the container
```

Working case:

```python
class AnyClass:
    pass

self.application.simple(AnyClass)

instance = self.application.make(AnyClass)

true == isinstance(instance, AnyClass)
```

```python
class AnyClass:
    pass

# never bound via self.application.bind/simple/etc
instance = self.application.make(AnyClass)

true == isinstance(instance, AnyClass)
```

_No response_

No

### Anything else ?

This would be backwards compatible with existing bindings since it would be implemented as a fallback.

Resolution would still fail if arguments are necessary when instantiating and they cannot be resolved (e.g., str, int, etc).

Resolvable dependencies in the `def __init__(...)` args within this "fallback" resolution should continue to be resolved.

os-nikita:
This PR adds the ability to resolve class dependencies from the container that have not been previously bound.

- [x] Add tests for simple and nested cases
- [x] Add test for follow-up resolution of same class
- [x] Add comments for new behaviour
- [x] Add documentation in docs repo: https://github.com/MasoniteFramework/docs/pull/200

Closes #618

### DIFF
--- a/src/masonite/container/container.py
+++ b/src/masonite/container/container.py
@@ -96,11 +96,15 @@ class Container:
         elif name in self.swaps:
             return self.swaps.get(name)
         elif inspect.isclass(name):
-            obj = self._find_obj(name)
-            self.fire_hook("make", name, obj)
-            if inspect.isclass(obj):
-                obj = self.resolve(obj, *arguments)
-            return obj
+            try:
+                obj = self._find_obj(name)
+                self.fire_hook("make", name, obj)
+                if inspect.isclass(obj):
+                    return self.resolve(obj, *arguments)
+            except MissingContainerBindingNotFound:
+                # todo: comment
+                self.simple(name)
+                return self.make(name, *arguments)
 
         raise MissingContainerBindingNotFound(
             "{0} key was not found in the container".format(name)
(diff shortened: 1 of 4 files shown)


os-nikita:
I've added tests, comments, and docs. I think this may be good for final review?

girardinsamuel:
I merged this accidentally but we will review this PR soon !
Thanks for you contribution 😉

PR #621, opened 2022-05-08
Add `Redis` Session driver

circulon:
This is an initial implementation of a driver for Redis backed sessions

Interested in your opinions on how I have implemented this
 
Tests are not built yet.... coming soon

### DIFF
--- a/src/masonite/drivers/session/RedisDriver.py
+++ b/src/masonite/drivers/session/RedisDriver.py
@@ -85,9 +85,9 @@ class RedisDriver(BaseDriver):
         if not session_id:
             request = self.application.make("request")
             session_id = request.cookie("SESSID")
-        ns = self.options.get("namespace", "")
-        ns += ":" if ns else ""
-        return f"{ns}session:{session_id}:"
+        namespace = self.options.get("namespace", "")
+        namespace += ":" if namespace else ""
+        return f"{namespace}session:{session_id}:"
 
     def get_data_prefix(self):
         return f"{self.get_session_namespace()}data:"
--- a/src/masonite/drivers/session/RedisDriver.py
+++ b/src/masonite/drivers/session/RedisDriver.py
@@ -6,7 +6,6 @@ from ...validation import MessageBag
 
 if TYPE_CHECKING:
     from ...foundation import Application
-    from ...request import Request
 
 
 class RedisDriver(BaseDriver):
@@ -15,6 +14,7 @@ class RedisDriver(BaseDriver):
     def __init__(self, application: "Application"):
         super().__init__(application)
         self.connection = None
+        self.driver_config = {"decode_responses": True}
 
     def get_connection(self):
         try:
@@ -25,10 +25,8 @@ class RedisDriver(BaseDriver):
             )
 
         if not self.connection:
-            driver_config = self.options['driver_config']
-            driver_config.update({"decode_responses": True})
             self.connection = redis.StrictRedis(
-                **driver_config,
+                **self.driver_config,
             )
 
         return self.connection
--- a/src/masonite/drivers/session/RedisDriver.py
+++ b/src/masonite/drivers/session/RedisDriver.py
@@ -1,5 +1,5 @@
 import json
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from redis import Redis
 
@@ -61,7 +61,7 @@ class RedisDriver(BaseDriver):
         return {"data": data, "flashed": flashed}
 
     def save(
-        self, added=None, deleted=None, flashed=None, deleted_flashed=None
+            self, added: dict = None, deleted: list = None, flashed: dict = None, deleted_flashed: list = None
     ) -> None:
         if added is None:
             added = {}
@@ -86,7 +86,7 @@ class RedisDriver(BaseDriver):
         for key in deleted_flashed:
             self.delete(f"{self.get_flash_prefix()}{key}")
 
-    def get_session_namespace(self, session_id=None) -> str:
+    def get_session_namespace(self, session_id: str = None) -> str:
         if not session_id:
             request = self.application.make("request")
             session_id = request.cookie("SESSID")
@@ -104,14 +104,14 @@ class RedisDriver(BaseDriver):
         """Use to create builtin helper function."""
         return self
 
-    def get(self, key, default=None, **options) -> str:
+    def get(self, key: str, default: str = None) -> str:
         if not self.has(key):
             return default
         return self.get_value(
             self.get_connection().get(key)
         )
 
-    def set(self, key, value, time: int = None):
+    def set(self, key: str, value: Any, time: int = None):
         time = time or self.get_timeout()
 
         if isinstance(value, (dict, list)):
@@ -119,13 +119,13 @@ class RedisDriver(BaseDriver):
 
         return self.get_connection().set(key, value, ex=time)
 
-    def has(self, key):
+    def has(self, key: str):
         return self.get_connection().get(key)
 
-    def delete(self, key):
+    def delete(self, key: str):
         return self.get_connection().delete(key)
 
-    def flush(self, session_id=None) -> None:
+    def flush(self, session_id: str = None) -> None:
         """
         Clears all data for the current (or provided) session id
         """
@@ -142,7 +142,7 @@ class RedisDriver(BaseDriver):
         # default timeout of session vars is 24 hrs
         return int(self.options.get("timeout", 60 * 60 * 24))
 
-    def get_value(self, value) -> str:
+    def get_value(self, value: Any) -> str:
         value = str(value)
         if value.isdigit():
             return str(value)
--- a/src/masonite/drivers/session/RedisDriver.py
+++ b/src/masonite/drivers/session/RedisDriver.py
@@ -51,9 +51,9 @@ class RedisDriver(BaseDriver):
         data_prefix = self.get_data_prefix()
         flash_prefix = self.get_flash_prefix()
         for key, value in session_data.items():
-            if key.startswith(data_prefix):
+            if str(key).startswith(data_prefix):
                 data.update({key.replace(data_prefix, ""): value})
-            elif key.startswith(flash_prefix):
+            elif str(key).startswith(flash_prefix):
                 flashed.update({key.replace(flash_prefix, ""): value})
 
         return {"data": data, "flashed": flashed}
--- a/src/masonite/drivers/session/RedisDriver.py
+++ b/src/masonite/drivers/session/RedisDriver.py
@@ -86,7 +86,7 @@ class RedisDriver(BaseDriver):
         for key in deleted_flashed:
             self.delete(f"{self.get_flash_prefix()}{key}")
 
-    def get_session_namespace(self, session_id=None):
+    def get_session_namespace(self, session_id=None) -> str:
         if not session_id:
             request = self.application.make("request")
             session_id = request.cookie("SESSID")
@@ -94,17 +94,17 @@ class RedisDriver(BaseDriver):
         namespace += ":" if namespace else ""
         return f"{namespace}session:{session_id}:"
 
-    def get_data_prefix(self):
+    def get_data_prefix(self) -> str:
         return f"{self.get_session_namespace()}data:"
 
-    def get_flash_prefix(self):
+    def get_flash_prefix(self) -> str:
         return f"{self.get_session_namespace()}flash:"
 
     def helper(self) -> "RedisDriver":
         """Use to create builtin helper function."""
         return self
 
-    def get(self, key, default=None, **options):
+    def get(self, key, default=None, **options) -> str:
         if not self.has(key):
             return default
         return self.get_value(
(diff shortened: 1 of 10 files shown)


girardinsamuel (commented):
For now the architecture looks good !

girardinsamuel:
Hi ! 
Looks like linting needs to be fixed on this PR 😉

circulon:
@girardinsamuel @josephmancuso 

OK I think this is completed now.
Added tests and reworked config and general compatibility with rxisting session methods

src/masonite/drivers/session/RedisDriver.py
```diff
@@ -0,0 +1,147 @@
...
+            self.delete(f"{self.get_data_prefix()}{key}")
+
+        for key in deleted_flashed:
+            self.delete(f"{self.get_flash_prefix()}{key}")
+
+    def get_session_namespace(self, session_id=None):
+        if not session_id:
+            request = self.application.make("request")
+            session_id = request.cookie("SESSID")
+        ns = self.options.get("namespace", "")
```
josephmancuso: ns variable should be called `namespace`. Remember code is read more than its written. No need for shorthanded variable names

src/masonite/sessions/Session.py
```diff
@@ -31,19 +34,31 @@ def driver(self, driver: str) -> Any:
 
     def set_configuration(self, config: dict) -> "Session":
         """Set session driver options."""
+
+        # make sure the default drivewr is defined
```
josephmancuso: typo. `drivewr` -> `driver`

tests/integrations/config/session.py
```diff
@@ -4,4 +4,13 @@
 DRIVERS = {
     "default": "cookie",
     "cookie": {},
+    "redis": {
+        "driver_config": {
```
josephmancuso: Why are we wrapping this inside a driver_config dictionary? Can' we just do `host`, `port`, `password` inside the `redis` connection? This looks ugly to me and not really similar to other features

josephmancuso (changes requested):
Looking good! I gave it a code review. Just some things I saw and some comments to make the code more similar to our other features

src/masonite/drivers/session/RedisDriver.py
circulon: Fixed

src/masonite/sessions/Session.py
circulon: Fixed

tests/integrations/config/session.py
circulon: The reason for this is that if you do this in the same way as per the Redis Cache Driver there is no easy way to provide additional initial configuration to the underlying connection driver StrictRedis.
 ie how do I set the "db=3" on the connection or enable any of the ssl options?
There is no facility for this at all nor on any of the other drivers that I could tell 
In the case of the StrictRedis driver there are 35+ options available to configure the driver.

This  might cause the developer for example to write a subclass of the RedisDriver to override the get_connection() method to use out of the box  options that the driver module provides. 

Yes it is a bit ugly I admit but I could not see a better way to describe the options that would be applied to the module driver itself.
While still providing configuration to the wrapper (session in this case) for things like timeout and namespace.  
Putting all module config and wrapper config options at the same level means you then have to separate them at the wrapper level by some other means.   

Ideally the framework should provide the ability to use the underlying module the ServiceProvider wraps with any configuration that is defined by the underlying module/s it uses in as easy a manner as possible. 
Not force the developer  to write additional code to handle something the underlying module (driver) already provides.

I hope this clarifies the design choice.

Happy to discuss more about better ways to provide this functionality

src/masonite/drivers/session/RedisDriver.py
```diff
@@ -0,0 +1,147 @@
...
+    def get_connection(self):
+        try:
+            import redis
+        except ImportError:
+            raise ModuleNotFoundError(
+                "Could not find the 'redis' library. Run 'pip install redis' to fix this."
+            )
+
+        if not self.connection:
+            self.connection = redis.StrictRedis(
```
girardinsamuel: On an other note maybe we could upgrade redis version and use Redis (https://github.com/redis/redis-py#client-classes-redis-and-strictredis) since 3.X StrictRedis has been renamed to Redis.

tests/integrations/config/session.py
girardinsamuel: Maybe the solution is to do something like this:

```python
DRIVERS = {
    #...
    "redis": {
        "host": "127.0.0.1",
        "port": "6379",
        "password": "",
        "options": {},
        "namespace": "",
        "timeout": "",
    }
}
```
and then build the connection like this:

```python
self.connection = Redis(
    **self.options.get("options"),
    host=self.options.get("host"),
    port=self.options.get("port"),
    password=self.options.get("password"),
)
```

(It's what is used for Masonite ORM config)

tests/features/session/test_cookie_session.py
```diff
@@ -6,42 +6,35 @@ def test_can_start_session(self):
         request = self.make_request()
         session = self.application.make("session")
         request.cookie("s_hello", "test")
-        session.start()
+        session.start("cookie")
```
girardinsamuel, line 9: Can you explain this change ? 🤔 I am not sure to follow.

girardinsamuel (changes requested):

tests/features/session/test_cookie_session.py
circulon: In this case although the "default" config for these tests is defined as "cookie" this may be changed by someone else when building other tests for other drivers.
This makes sure we are explicitly testing the "cookie" driver and not relying on the "default" which may or may not be "cookie"

src/masonite/drivers/session/RedisDriver.py
circulon: Yeah thats makes sense and keeps Masonite up to date with the backing module 

> On an other note maybe we could upgrade redis version and use Redis (https://github.com/redis/redis-py#client-classes-redis-and-strictredis) since 3.X StrictRedis has been renamed to Redis.

That work for me and keeps Masonite inline with the backend module

tests/features/session/test_cookie_session.py
circulon: Re: the driver config ... I'm still learning the nuances of Python re passing options that have not actually assignments for the class.  
ie I assumed (incorrectly) that the class init would complain that the "timeout" and "namespace" kwargs which are not actually Redis() (or StrictRedis()) args would not work.

so yeah if this works I'm fine with it and so it will align with the other configs and reduce complexity.  

Will report back if this doesnt work as expected.

circulon: @girardinsamuel 
after rereading your "options" config again this make much more sense... sorry its been a long day and I'm tired ;)  

so Yes the "options" at the root of the driver config makes much more sense and is a much cleaner option...
Will test this scenario in the next couple of days or so.  
Cheers for the pointers

tests/integrations/config/session.py
circulon: @girardinsamuel  thanks thats much cleaner and still provides flexibility.
Implemented it and its working ;)

josephmancuso (approved):

girardinsamuel:
Hi @circulon, thank you again for your contribution.
One thing I missed in the first review are the type hints. We are progressively type hinting the whole Masonite codebase (It should be completed for Masonite 5). As the Cookie session driver is type hinted, could you do the same for Redis session driver ? You can check how it's done in `masonite/drivers/session/CookieDriver.py`. 👍 

Note: i have seen it's partially done, so it should just be adding some missing type hints on some methods.

circulon:
> Hi @circulon, thank you again for your contribution. One thing I missed in the first review are the type hints. We are progressively type hinting the whole Masonite codebase (It should be completed for Masonite 5). As the Cookie session driver is type hinted, could you do the same for Redis session driver ? You can check how it's done in `masonite/drivers/session/CookieDriver.py`. 👍
> 
> Note: i have seen it's partially done, so it should just be adding some missing type hints on some methods.

Hi @girardinsamuel 
I appreciate the thanks ;)

Yeah my project requires Redis for the cache and I use a thitf party auth system so internal lookups to validate the user token are just not feasible.
So I decided to put the user object in the session as many other frameworks do.  Turns out having tokens in the user object exceeds the cookie size limit ;(  

As the Redis Cache driver was already built it was just a matter of reworking that driver to fit the way Sessions work.
 So there is a lot of reused code from the cache driver as you might have noticed.

Regarding Type hinting I have added most of them.  There are a couple of methods like "set" that I'm not exactly sure of the return type so any pointers would be appreciated.

tests/integrations/config/session.py
```diff
@@ -5,11 +5,8 @@
     "default": "cookie",
     "cookie": {},
     "redis": {
-        "driver_config": {
-            "host": "127.0.0.1",
-            "port": "6379",
-            "password": "",
-        },
+        "host": "127.0.0.1",
```
girardinsamuel, line 8: I know you put default values inside Redis session driver connection, but could you in addition put "port" and "password" here again ? It will help users wanting to use this driver to configure it as they will directly see those base options. 👍 Also it will more consistent with options of redis cache driver.

src/masonite/drivers/session/RedisDriver.py
```diff
@@ -81,25 +86,25 @@ def save(
         for key in deleted_flashed:
             self.delete(f"{self.get_flash_prefix()}{key}")
 
-    def get_session_namespace(self, session_id=None):
+    def get_session_namespace(self, session_id=None) -> str:
```
girardinsamuel:
```suggestion
    def get_session_namespace(self, session_id: str = None) -> str:
```

src/masonite/drivers/session/RedisDriver.py
```diff
@@ -81,25 +86,25 @@ def save(
...
-    def get_flash_prefix(self):
+    def get_flash_prefix(self) -> str:
         return f"{self.get_session_namespace()}flash:"
 
     def helper(self) -> "RedisDriver":
         """Use to create builtin helper function."""
         return self
 
-    def get(self, key, default=None, **options):
+    def get(self, key, default=None, **options) -> str:
```
girardinsamuel:
```suggestion
    def get(self, key: str, default: Any = None, **options) -> str:
```

src/masonite/drivers/session/RedisDriver.py
```diff
@@ -1,6 +1,8 @@
 import json
 from typing import TYPE_CHECKING
```
girardinsamuel:
```suggestion
from typing import TYPE_CHECKING, Any
```

src/masonite/drivers/session/RedisDriver.py
```diff
@@ -120,9 +125,9 @@ def has(self, key):
     def delete(self, key):
```
girardinsamuel:
```suggestion
    def delete(self, key: str):
```

src/masonite/drivers/session/RedisDriver.py
```diff
@@ -120,9 +125,9 @@ def has(self, key):
...
         return self.get_connection().delete(key)
 
-    def flush(self, session_id=None):
+    def flush(self, session_id=None) -> None:
```
girardinsamuel:
```suggestion
    def flush(self, session_id: str = None) -> None:
```

src/masonite/drivers/session/RedisDriver.py
```diff
@@ -133,11 +138,11 @@ def flush(self, session_id=None):
...
                 connection.delete(*keys)
 
-    def get_timeout(self):
+    def get_timeout(self) -> int:
         # default timeout of session vars is 24 hrs
-        return self.options.get("timeout", 60 * 60 * 24)
+        return int(self.options.get("timeout", 60 * 60 * 24))
 
-    def get_value(self, value):
+    def get_value(self, value) -> str:
```
girardinsamuel:
```suggestion
    def get_value(self, value: Any) -> str:
```

src/masonite/drivers/session/RedisDriver.py
girardinsamuel: For set() you can type hint like this:
```python
def set(self, key: str, value: Any, time: int = None):
```

For has() you can type hint like this:
```python
def has(self, key: str) -> bool:
```

girardinsamuel (commented):
Here are missing type hints to help you.
For some return types you have to look at redis methods return type to know what is the type.

tests/integrations/config/session.py
circulon: Yeah Makes sense...
Done

src/masonite/drivers/session/RedisDriver.py
circulon: Aah yes I see what you mean ok ....  Done

src/masonite/drivers/session/RedisDriver.py
circulon: Yeah I see ... For some reason I thought "type hinting" meant the hint for the return value of methods.
I learnt something new today... 
Cheers

Done

girardinsamuel (approved):
Final review ! Looks good to me.
@josephmancuso I guess we can merge this now.

josephmancuso (approved):
Final code review looks good

PR #628, opened 2022-05-15
Adding support to define custom region in mailgun endpoint.

### LINKED ISSUE #627: Mailgun Driver does not support `eu` domains in API
### Describe the bug

When adding a new domain in Mailgun, you have the option of specifying EU as the region; US is the default. If the developer does so, then the base API url changes from `https://api.mailgun.net/v3/` to `https://api.eu.mailgun.net/v3/`.

The Masonite Mailgun driver is currently hardcoded to the US endpoint with no way of supporting the EU endpoint without adding a custom driver.

image

### Expected behaviour

_No response_

_No response_

### Screenshots

_No response_

### OS

macOS

### OS version

Monterey 12.3.1

### Browser

_No response_

### Masonite Version

4.12.1

### Anything else ?

_No response_

Marlysson:
Closes #627

### DIFF
--- a/src/masonite/mail/drivers/MailgunDriver.py
+++ b/src/masonite/mail/drivers/MailgunDriver.py
@@ -43,11 +43,19 @@ class MailgunDriver:
 
     def send(self):
         domain = self.options["domain"]
+        region = self.options["region"]
         secret = self.options["secret"]
         attachments = self.get_attachments()
 
+        BASE_URL_BY_REGION = {
+            "us": "https://api.mailgun.net/v3/",
+            "eu": "https://api.eu.mailgun.net/v3/"
+        }
+
+        endpoint = BASE_URL_BY_REGION.get(region.lower(), default=BASE_URL_BY_REGION.get("us"))
+
         return requests.post(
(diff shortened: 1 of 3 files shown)


girardinsamuel (changes requested):
It looks good to me.

Just a minor change:
- Can you add "region" key, default to "us" in tests/integrations/config/mail.py ?
- Can you open a PR in cookiecutter repo to add the config key too ?

src/masonite/mail/drivers/MailgunDriver.py
```diff
@@ -43,11 +43,19 @@ def get_attachments(self):
 
     def send(self):
         domain = self.options["domain"]
+        region = self.options["region"]
```
josephmancuso: I don't know where my comment went but we need to change this to a `self.options.get("region", "us")`. This is because the region is not going to be set on some people configs

Marlysson:
- [x] Can you add "region" key, default to "us" in tests/integrations/config/mail.py ?
- [x] Can you open a PR in cookiecutter repo to add the config key too ?
- [x] self.options.get("region", "us")

josephmancuso (approved):

PR #629, opened 2022-05-15
fixed auth

josephmancuso:


girardinsamuel:
I also wanted to do this, but I have found that this was causing tests to fail...
I am not sure yet what they fail, we have to look into it in details I guess

girardinsamuel:
Actually it's normal that tests are failing and I am not sure that the correct behaviour for real life too...

- RouteProvider is placed before AuthenticationProvider
- With `RouteProvider.boot()`,  the controller method will be ran first (and will authenticate the user), after we have a user logged in
- Then `AuthenticationProvider.boot()` will be ran logging out the user
- So at the end of the request lifecycle the user is always logged out.

It is not kinda weird ?

Should not Auth reset at start of request instead of resetting it at the end ?

josephmancuso:
I added a test guard here to fix testing issue

I think the next thing here is to be able to test actual guards like the web guard

josephmancuso (commented):
reviewed

girardinsamuel:
I am feeling that adding a TestGuard is a bit weird...instead of using the same auth mechanism in tests we are creating an other one ? I find that a bit worrying as we won't be able to test the auth guards correctly IMO

josephmancuso:
@girardinsamuel we can't test guards right now. I'll need to fix this to fix the tests and then I'll revisit the testing of the actual guards. Its going to be pretty tricky I think

girardinsamuel:
> @girardinsamuel we can't test guards right now. I'll need to fix this to fix the tests and then I'll revisit the testing of the actual guards. Its going to be pretty tricky I think

I tried it and I was very close. The issue was explained here:
https://github.com/MasoniteFramework/masonite/pull/629#issuecomment-1129796151

Issue #630, opened 2022-05-16
Change how Validation rules are configured


circulon:
### Is your feature request related to a problem?

Currently in 4.0 the Dictionary method for defining multiple field validation rules and customising them is not as easy as it could be.
Some customisations will first require PR's to fix the validator before it can be used with custom options. 
This would be a less preferable method as it will not address the underlying customisation issues

Example:
In the RegistrationController that is built as part of the craft command the 'store' method defines the basic field validation as similar to this:
```python
errors = request.validate(
            {
                "email": "required|email",
                "password": "required|strong|confirmed",
            }
        )
``` 
This is fairly straight forward but if you want to customise the password complexity like this:
```pthhon
...
 "password": "required|strong:length=10,uppercase=1,special=3|confirmed",
...
```
you will get an exception that 'str' has no '<' operator due to the value that is assigned to the attributes being part of a string.
This not ideal.

This is beside the point of this feature request but it illustrates that the configuration of the validator ("strong" in this case) is not that obvious or easily understood.  
I actually had to read through the code to understand (then fail with an exception) and customise the validators criteria.

### What do we currently have to do now?

As per the example above a PR would need to be created to cast the assigned attributes to 'int' before comparison.

### Describe the solution you'd like

What I would propose is that we change the way validators are configured to be more pythonic and play nice with ide's ie context hilighting.

Based on my experiences using Yii Validation Rules
https://www.yiiframework.com/doc/guide/2.0/en/input-validation#declaring-rules

Something like the following would be more flexible and obvious when writing and reviewing these configs, additionally the attributes passed to the validator when customising will retain their type.

```python
errors = request.validate([
  (["email", "password", "terms_accepted"], "required"),
  (["email"], "email"),
  (["password"], "strong", {length=10, uppercase=1, special=3}),
  (["password"], "confirmed"),
  (["terms_accepted"], "accepted"),
])
```
The upshot of this is that the valdator method accepts a list of tuples where each tuple describes a validation rule context/s.

This also means that the need fpr grouping rules in a RuleEnclosure may become less as the ease of customisation increses.

This would not need to be a breaking chang but instead a "DeprecationWarning" for the legacy dictionary method could be generated. 

### Describe alternatives you've considered

None that I can think of but happy to discuss

### Would this be a breaking change ?

- [ ] Yes

### Anything else ?

I am seriously considering putting together a PR for this to ease my rule customisation woes.

josephmancuso:
[...]

PR #632, opened 2022-05-17
change and simplify validation rules config

circulon:
This PR addresses #630 providing a simplified and consistent interface for configuring per call field validation criteria


josephmancuso:
Before you put more work on this I think we should discuss this more. I'm not yet sold on this and if there is a huge API change to the point where you are trying to manage a legacy format and a new format then maybe we just do the new format and put it into a 5.x branch.

With that being said though, I'm not sure I am sold yet on the new format when we can do something similar already with lists and combining strings and functions

circulon:
Closing this as it is not a deal breaker thanks to the comments in #630

PR #634, opened 2022-05-17
Fixed exceptionite possible solutions for `InvalidCSRFToken` and `ModelNotFoundException`

### LINKED ISSUE #45: Fix solutions for Invalid CSRF token
image

The solution description has not been written.

### LINKED ISSUE #49: Add exceptionite solution for `ModelNotFoundException`
A solution is missing:
Exception namespace: `src.masonite.exceptions.exceptions.ModelNotFoundException`

girardinsamuel:
Fix https://github.com/MasoniteFramework/exceptionite/issues/49
Fix https://github.com/MasoniteFramework/exceptionite/issues/45

### DIFF
--- a/src/masonite/exceptions/exceptionite/solutions.py
+++ b/src/masonite/exceptions/exceptionite/solutions.py
@@ -22,10 +22,10 @@ class MissingCSRFToken:
 
 class InvalidCSRFToken:
     def title(self):
-        return "Invalid CSRF Token"
+        return "The session does not match the CSRF token"
 
     def description(self):
-        return "The CSRF tokens found in the user session and in the request do not match. This is flagged as a potential CSRF attack."
+        return "Try clearing your cookies for the localhost domain in your browsers developer tools."
 
     def regex(self):
         return r"Invalid CSRF Token"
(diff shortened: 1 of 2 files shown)


girardinsamuel:
@josephmancuso I let you suggest a better message for invalid csrf token maybe ?

src/masonite/exceptions/exceptionite/solutions.py
```diff
@@ -25,7 +25,7 @@ def title(self):
         return "Invalid CSRF Token"
 
     def description(self):
-        return "TODO"
+        return "The CSRF tokens found in the user session and in the request do not match. This is flagged as a potential CSRF attack."
```
josephmancuso: Remember that solutions are solutions and should not reiterate the issue. They know what the issue with our detailed descriptions of this issue in the exception. "Invalid CSRF Token" or something.

Solutions should be actions the user should take to resolve the issue.

So the title should not be `Invalid CSRF Token` (we know that already from the exception) but rather something like:

Title: The session does not match the CSRF token
Description: Try clearing your cookies for the localhost domain in your browsers developer tools.

girardinsamuel: Hmm that's totally right !

PR #637, opened 2022-05-19
Add more tests for Container

girardinsamuel:
- Add more tests to Container
- Show the issue with the namespace. Instead of raising an error that the binding cannot be found an unhandled error is raised.

This also outlines that the Container is considering differently 
`src.masonite.response.Response` and `masonite.response.Response`.

### DIFF
--- a/src/masonite/container/container.py
+++ b/src/masonite/container/container.py
@@ -181,7 +181,7 @@ class Container:
                 raise ContainerError(str(e))
         else:
             for _, value in self.get_parameters(obj):
-                if value.annotation in (str, int, dict, list, tuple):
+                if type(value.annotation) in (str, int, dict, list, tuple):
                     # Ignore any times a user is simply type hinting a parameter like (parameter:str).
                     # In this case we don't want to resolve anything but we do want
                     # to insert any passing arguments we passed in
@@ -199,7 +199,11 @@ class Container:
                         # See `self.make()`.
                         param = value.annotation
                     if inspect.isclass(param):
-                        param = self.resolve(param)
+                        print(resolving_arguments)
+                        if resolving_arguments:
+                            param = self.resolve(param, *resolving_arguments)
+                        else:
+                            param = self.resolve(param)
                     objects.append(param)
                 elif "self" in str(value):
                     objects.append(obj)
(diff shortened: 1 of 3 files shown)


girardinsamuel:
If you look closely, it fails when trying to resolve the nested dependency (`Application`) passed in the init of the `SomeStringTypeHintedAppObject`. When type hinted with a string it causes an issue because when resolving the dependency it won't be able to find the class associated to "Application".

Although `Response` class is written the same way, as it is bound explicitly, the container never have to instantiate again on the fly and resolve this application parameter. That is why it's working with Response class and not with SomeStringTypeHintedAppObject class.

So currently, the "resolving on the fly" feature does not handle the string type hinted parameters.

src/masonite/container/container.py
```diff
@@ -292,14 +292,15 @@ def _find_annotated_parameter(self, parameter):
         Returns:
             object -- Returns the object found in the container.
         """
+        if isinstance(parameter.annotation, str):
+            return
```
josephmancuso, line 299: All i did was add this. This is good yeah?

girardinsamuel: Well it does solve the test, but then dependencies which are string type hinted and which are resolved on the fly, are going to be `None`. (I pushed a commit to demonstrate it)

It might not be an issue but it makes those changes a bit inconsistent...

src/masonite/container/container.py
```diff
@@ -181,7 +181,7 @@ def resolve(self, obj, *resolving_arguments):
                 raise ContainerError(str(e))
         else:
             for _, value in self.get_parameters(obj):
-                if value.annotation in (str, int, dict, list, tuple):
+                if type(value.annotation) in (str, int, dict, list, tuple):
```
josephmancuso, line 184: This entire problem was due to this line right here lol. This line i guess never worked. value.annotation was something like `"app: Application"` (a string). So that value was never in this list of types lol. We had to do the type(value.annotation) in order to check the type against the list of types. Changing this fixed the issue.

We now get an expected error like "__init__() missing required parameter app"

girardinsamuel: Good news !

girardinsamuel:
Looks good to be merged !

PR #639, opened 2022-05-22
Added explicit type casting for route parameters

### LINKED ISSUE #635: Add explicit type casting for route parameters
Type casting will not be achieved by route compilers as it's not its responsibility (cf Joe's comment below).

However, it would be great to be able to define how route parameters should be cast back to controller/request parameters.

## Specs
It could be done explicitly like this:

```python
Route.get("/user/@id").casts({"id": "int"})
```

or more quickly like this if the route compiler is named after a python type:
```python
Route.get("/user/@id:int").casts()
```

And then when using `request.param("id")` or resolving id from the controller `def details(self, request:Request, id)` we would get an integer.

girardinsamuel:
Fix #635.

### DIFF
--- a/src/masonite/routes/HTTPRoute.py
+++ b/src/masonite/routes/HTTPRoute.py
@@ -1,4 +1,3 @@
-from inspect import isclass
 import re
 from urllib import parse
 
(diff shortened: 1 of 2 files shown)


josephmancuso (approved):
first review looks good

josephmancuso:
Is there anything else? I think that's it right?

girardinsamuel:
I think too 😅

PR #653, opened 2022-06-02
Fix session drivers default options ignored

circulon:
This fixes the issue where default options for the driver were ignored  when setting  the options

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,3 +22,4 @@ tldextract>=2.2,<2.3
 waitress>=2.1,<2.2
 whitenoise>=5.2,<5.3
 pyjwt>=2.3,<2.5
+mergedeep>=1.3,<2.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,4 +22,3 @@ tldextract>=2.2,<2.3
 waitress>=2.1,<2.2
 whitenoise>=5.2,<5.3
 pyjwt>=2.3,<2.5
-mergedeep>=1.3,<2.0
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,6 @@ setup(
         "pytest>=7<8",
         "werkzeug>=2<3",
         "watchdog>=2<3",
-        "mergedeep>1,<2",
     ],
     # See https://pypi.python.org/pypi?%3Aaction=list_classifiers
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,7 @@ setup(
         "pytest>=7<8",
         "werkzeug>=2<3",
         "watchdog>=2<3",
+        "mergedeep>1,<2",
     ],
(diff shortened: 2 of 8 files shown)


josephmancuso:
Do we need a new package just to update dictionaries?

circulon:
> Do we need a new package just to update dictionaries?

Hi @josephmancuso 
As it stands using `.update()` on a dict will overwtite any key with the contents of the new value for that key.  
This means that If that key (in the original dict) contains a dict with values already defined, the new dict  will replace the old one regardless of what the old one contains. 
The `mergedeep` package allows combining nested multilevel dicts to produce a complete representation of all sources (keys and values) at all levels by merging the values (dict) and not replacing them.

Doing it this way also allows defaults to be set for a components/driver while still allowing the defaults to be altered if necessary in a very clean and obvious way. 

If there is another way to combine multilevel nested dicts without using for loops that lock you into a max nested level then I'm happy to discuss and work on this.

I hope this clarifies things.
Cheers

girardinsamuel:
The only "issue" I see here is that we did not implement default options in any Masonite drivers. All config is explicitly coming from the configuration files located in `config/`. I feel that putting default values here, hidden from the "end user" will create confusion... @josephmancuso what's your opinion on this ?

circulon:
> The only "issue" I see here is that we did not implement default options in any Masonite > drivers. All config is explicitly coming from the configuration files located in config/. I feel > that putting default values here, hidden from the "end user" will create confusion... 
> @josephmancuso what's your opinion on this ?
 
@girardinsamuel 
Actually there is default configuration set but it is hardcoded in many places.
Please note when I say "default configuration" I mean optional but quite often recommended.
eg in the `cache/drivers/RedisDriver.py`
```python
self.connection = redis.StrictRedis(
                host=self.options.get("host"),
                port=self.options.get("port"),
                password=self.options.get("password"),
                decode_responses=True,
            )
```

Note the `decode_responses` is hardcoded to `True` here.
This option for the redis session driver can be overridden if required allowing maximum flexibility. 

With regards to separated default_options yeah I do see that its a kind of anti pattern as the framework does not implement this kind of option separation.
It was just that as it was set in the `__init__` it felt more obvious. 

I will change it back to just use the `options` and add the default options in there instead for standards purposes.

Regards

circulon:
is this complete now?  I have addressed the default options issue

Thanks

circulon:
After thinking about this "having defaults that are adjustable" I realised that the `decode_responses` should not be a configurable option as the Session requires data it understands to store and retrieve internally.

So by allowing the Redis driver to return a raw  response this will break the Session Driver because it does not understand the raw Redis object.

I have now removed the `mergeedeep` dependency and set the
`decode_responses=True` on the `RedisDriver`

Apologies for my misunderstanding of what I was trying to achieve.
Cheers for the discussion too as it made me reexamine what I was actually doing.

josephmancuso (approved):

PR #655, opened 2022-06-02
Fixed providing application instance to listener constructors

### LINKED ISSUE #651: application instance is not provided to event listeners
### Describe the bug

It gives below error while trying to make custom exception handler using [Catch All Exceptions](https://docs.masoniteproject.com/the-basics/error-handling#catch-all-exceptions)

> Traceback (most recent call last):
  File "/Users/test/workspace_py/test/examples/masonite/mas_venv/lib/python3.9/site-packages/werkzeug/serving.py", line 335, in run_wsgi
    execute(self.server.app)
  File "/Users/test/workspace_py/test/examples/masonite/mas_venv/lib/python3.9/site-packages/werkzeug/serving.py", line 322, in execute
    application_iter = app(environ, start_response)
  File "/Users/test/workspace_py/test/examples/masonite/mas_venv/lib/python3.9/site-packages/masonite/foundation/Application.py", line 61, in __call__
    return self.response_handler(*args, **kwargs)
  File "/Users/test/workspace_py/test/examples/masonite/mas_venv/lib/python3.9/site-packages/whitenoise/base.py", line 85, in __call__
    return self.application(environ, start_response)
  File "/Users/test/workspace_py/test/examples/masonite/mas_venv/lib/python3.9/site-packages/masonite/foundation/response_handler.py", line 39, in response_handler
    application.make("exception_handler").handle(e)
  File "/Users/test/workspace_py/test/examples/masonite/mas_venv/lib/python3.9/site-packages/masonite/exceptions/ExceptionHandler.py", line 34, in handle
    self.application.make("event").fire(
  File "/Users/test/workspace_py/test/examples/masonite/mas_venv/lib/python3.9/site-packages/masonite/events/Event.py", line 38, in fire
    listener().handle(event, *args, **kwargs)
TypeError: __init__() missing 1 required positional argument: 'application'

### Expected behaviour

While the event is firing, the application object should pass.

Simply follow the instructions in the [documentation](https://docs.masoniteproject.com/the-basics/error-handling#catch-all-exceptions). The ExceptionListener class is found in another package, which is imported and registered in 'AppProvider.py'.

### Screenshots

_No response_

### OS

macOS

### OS version

Monterey 12.2

### Browser

_No response_

### Masonite Version

4.14.1

### Anything else ?

_No response_

girardinsamuel:
This PR fixes #651 allowing to access app instance inside a listener.

⚠️ For now this is a breaking change for users which have written listeners without the `__init__(self, application)` method...
Either we add a check to pass the application instance or not for M4 (that we would remove in M5), or we release this in M5 only.

### DIFF
--- a/src/masonite/events/Event.py
+++ b/src/masonite/events/Event.py
@@ -1,5 +1,4 @@
 """ Event Module """
-
 import inspect
 
 
@@ -32,7 +31,7 @@ class Event:
             collected_events = self.collect_events(event)
             for collected_event in collected_events:
                 for listener in self.events.get(collected_event, []):
-                    listener().handle(event, *args, **kwargs)
+                    listener(self.application).handle(event, *args, **kwargs)
             return collected_events
         else:
             if inspect.isclass(event):
@@ -41,7 +40,7 @@ class Event:
             else:
                 lookup = event.__class__
             for listener in self.events.get(lookup, []):
-                listener().handle(event, *args, **kwargs)
+                listener(self.application).handle(event, *args, **kwargs)
 
             return [event]
 
--- a/src/masonite/stubs/events/Listener.py
+++ b/src/masonite/stubs/events/Listener.py
@@ -1,3 +1,6 @@
 class __class__:
-    def handle(self, event):
+    def __init__(self, application):
+        self.application = application
+
+    def handle(self, event, *args, **kwargs):
         pass
--- a/tests/features/event/test_event.py
+++ b/tests/features/event/test_event.py
@@ -23,21 +23,33 @@ class NewUserEvent(Event):
 
 
 class SendEmailListener:
+    def __init__(self, application):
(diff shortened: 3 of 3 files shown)


josephmancuso:
> For now this is a breaking change for users which have written listeners without the __init__(self, application) method...

I think thats OK because its not the documented way to do this anyway so they shouldn't be doing that

josephmancuso:
You know .... now that I look at this in real life i don't think it should be done like this lol.

1. Needing to add the application class on these listeners leads to an ugly interface
2. In the future when we start to queue the listeners, we're not going to be able to because the container cannot be pickled an serialized.

Maybe we resolve the listener? or we update the documentation to need to fetch the container like we said in the issue

josephmancuso:
Let's just update back the documentation to fetch the container through an import

PR #686, opened 2022-06-27
Redis Cache - Use internal cache for performance & bug fixes

circulon:
Use internal cache for Redis Driver 
Fixes #694 in Redis Driver 
Supercedes #684


src/masonite/cache/drivers/RedisDriver.py
```diff
@@ -1,91 +1,139 @@
 import json
+from typing import Any
+
+try:
```
girardinsamuel: You should keep (as before) the lazy loaded import here so that users can use Masonite without installing Redis.
I know that in order to type hint the code you need the Redis class imported on top, but here you should rely instead on `TYPE_CHECKING` flag to do this 👇 

```python
from typing import Any, TYPE_CHECKING

if TYPE_CHECKING:
    from redis import Redis
```

circulon: Makes sense yeah.  
Always the way right when your just looking at from your use case ;)

Fixed now

josephmancuso (approved):

PR #703, opened 2022-08-15
Fixed loading cookies on Linux environments

BooStudio:
Hello,

I came across this issue which I could not explain, but my best guess is it's got something to do with the Linux,
where the `self.environ.get("HTTP_COOKIE", "")` returns the cookie_string which is delimited with ";" whereas on my local Mac environment it returns the cookie_string which is delimited with "; " which creates an issue where the API calls fail as the CSRF tokens doesn't match.

The solution provided in the PR is a better way to handle cookies it's inspired by how Django & Laravel handles cookies and should work for both scenarios.

### DIFF
--- a/src/masonite/cookies/CookieJar.py
+++ b/src/masonite/cookies/CookieJar.py
@@ -59,9 +59,10 @@ class CookieJar:
         return dic
 
     def load(self, cookie_string):
-        for compound_value in cookie_string.split("; "):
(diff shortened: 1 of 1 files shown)


josephmancuso:
how was this PR open? Its not running any of the tests for some reason

josephmancuso (approved):
this looks fine though

PR #718, opened 2022-10-10
[5.X] Add `logging` feature

### LINKED ISSUE #663: Add `logging` feature with different logging handlers

Kind of.
Currently there is no way to log any form of message (exceptions, errors or debug info to persistent storage for later evaluation.

This is a problem in production scenarios where you want to store the message for evaluation and bug reproduction/confirmation.

There is no easy out of the box way to do this.

Possibly a `LoggingProvider` that can register handler/s 
Each  handler listens for (or processes only) specific log level messages in a specific way 

eg
Each handler would receive the message as by the `logging` module (eg logging.info()).
The handler would then process the message in whatever way is designed for.
ie send to a file,  add a row in a db table, custom handler to send to an external service via api, etc.

 This offers maximum flexibility and configuration for logging messages of any type regardless of environment or deployment strategy.

Is there something I have missed with this? 

- [ ] Yes

### Anything else ?

_No response_

girardinsamuel:
Fix MasoniteFramework/masonite#663.

# Specifications

Available drivers are: `terminal`, `single`, `daily`, `syslog`, `slack`.
Logging configuration is done in `config/logging.py`

```python
CHANNELS = {
    "default": {
        "driver": "console",
        # those settings will be applied to each channel if not overriden
        "level": "info", 
        "timezone": "UTC",
        "format": "{timestamp} - {levelname}: {message}",
        "date_format": "YYYY-MM-DD HH:mm:ss",
    },
    "console": {
        "driver": "terminal",
    },
    "single": {
        "driver": "single",
        "path": "logs/single.log",
    },
    "daily": {
        "driver": "daily",
        "path": "logs/daily.log",
        "days": 7,
        "keep": 10,
    },
    "all": {"driver": "stack", "channels": ["single", "daily", "console"]},
    "syslog": {"driver": "syslog", "address": "/var/log/system.log"},
    "papertrail": {
        "driver": "syslog",
        "host": "logs.papertrailapp.com",
        "port": None,  # specify here the port as an integer
    },
    "slack": {
        "driver": "slack",
        "webhook_url": "",
    },
}
```

```python
from masonite.facades import Log

# Log a message to the default configured channel
Log.debug("hello !")
Log.info("hello !")
Log.notice("hello !")
Log.warning("hello !")
Log.critical("hello !")
Log.critical("hello !")
Log.alert("hello !")
Log.emergency("hello !")

# You can change channel on the fly
Log.channel("daily").info("Some message")

# You can stack channels to log a message to multiple channels
Log.stack("console", "daily", "slack").critical("Some message")
```

Actually Sentry can be used in two different ways. One can add a custom exception handler and use [sentry python binding](https://github.com/getsentry/sentry-python) to send the info or there could be a Sentry logging driver, but it would not do much more than the python sentry binding. So what I propose is to maybe create a package for integrating a sentry exception handler a bit like [sentry-laravel](https://github.com/getsentry/sentry-laravel).

# What's left to do

- [x] add `syslog`, `slack` drivers.
- [ ] ensure exceptions are logged with enough information -> needs to work a bit ExceptionHandler code
- [ ] add `propagate` option
- [ ] add documentation https://github.com/MasoniteFramework/docs/pull/221
- [x] add catch all listener to log exceptions


BSN4:
great feature and great code looks awesome💥

circulon:
Nice work @girardinsamuel 

This looks exactly like what I was getting at in #663.
Functional and flexible.

Keep up the great work.
Cheers

PR #722, opened 2022-10-18
Fix incorrect extension being returned

cercos:
When a filename has multiple dots it returns everything after first dot.  os.path.splitext() correctly handles dot names and even '.hidden_linux_file'

### DIFF
--- a/src/masonite/utils/filesystem.py
+++ b/src/masonite/utils/filesystem.py
@@ -87,4 +87,3 @@ def get_extension(filepath: str, without_dot=False) -> str:
     if without_dot:
         return extension[1:]
     return extension
-
--- a/src/masonite/utils/filesystem.py
+++ b/src/masonite/utils/filesystem.py
@@ -76,7 +76,7 @@ def get_module_dir(module_file):
 def get_extension(filepath: str, without_dot=False) -> str:
     """Get file extension from a filepath. If without_dot=True the . prefix will be removed from
     the extension."""
-    extension = "".join(pathlib.Path(filepath).suffixes)
+    extension = os.path.splitext(filepath)[1]
     if without_dot:
         return extension[1:]
     return extension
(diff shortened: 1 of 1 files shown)


girardinsamuel:
Thanks for this PR. Can you add a unit test to demonstrate that it solves the issue ? 

I will review this asap !

cercos:
I will try.   I have never created one but I'm interested in the process.  Are unit tests universal or do frameworks integrate them in a specific way and is there a specific package I should be using for masonite?

girardinsamuel:
We are using `pytest` to run Masonite test suite (tests are based on `unittest`). There are some handy features available to write Masonite unit tests: https://docs.masoniteproject.com/testing/getting-started.

Here it is easy because tests already exists for this method, so you can edit it: tests/core/utils/test_filesystem.py.

It demonstrates that it works with multiple dots in the extension...so what extension did you tried that did not work here ? 🤔

cercos:
Ok thank you I will read up, didnt even think of looking at the docs lol.  The exact filename I was using to test the upload was "archlinux-2022.09.03-x86_64.iso" and it returns ".09.03-x86_64.iso" as the extension.

girardinsamuel:
I might have found a better solution (more automated) avoiding users to fill an env variable.
I am still using `pathlib` as it's doing a good job at finding all part of the name which contains `.` and if the last one is not a known extension I am joining all parts to build an extension.

I have added more tests to show that it solves your problem. If you have tricky extensions in mind feel free to add them to PR comments and I will add them to unit tests to check if this implementation can handle them too...

You can check it out here: https://github.com/MasoniteFramework/masonite/pull/727

josephmancuso:
Does #727 succeed this PR?

girardinsamuel:
> Does #727 succeed this PR?

Yes problem solved 💪 !

PR #751, opened 2023-04-02
Feature: add Amazon Services Provider

circulon:
This Provider provides a nice uniform way of accessing the low level 'client' or 'resource' interfaces provided by boto3 to AWS services.
In addition it provides a way to consolidate (service and app) configuration  for each service to use in your app.

Use it like this

- Add the `AmazonServicesProvider` to your `providers.py`
- Add a services config `amazon.py` containing services you want to use in your app
- Use the AmazonService class in your app

NOTE: the resources interface is still available but no longer being updated with new services.  
It is recommended by Amazon to use the client interface for access to all services.
A Deprecation warning will be raised if the 'resource' interface is accessed.  

```python
#config/amazon.py
SERVICES = {
    "s3": {
        # all keys except "service", "active" and "options" are useful for app specific configuration
        "buckets": {
            "images": "the-images",
            "console": "console-attachments",
        },
        "active": True, # Optional: defaults to True if not provided
        # any keys in the 'options' will be passed directly to the client or resource interface
        "options": { 
            "endpoint_url": "http://localhost:4566", # This is useful for development with LocalStack
        },
    },
    # using a more readable alias for the service
    "api_gateway": {
        "service": "apigateway", # this is the service name that boto3 uses
    },
}

# OPTIONAL: configuration that is common to all services
# NOTE: this can be overridden in the service `options` block if necessary 
COMMON_CONFIG = {
    "aws_access_key_id": "<access key ID>",
    "aws_secret_access_key": "<secret key>",
    "region_name": 'us-eat-1',
}

```
In your Controller 
```python
# app/controllers/SomeController.py

from masonite.services import AmazonServices
...
class SomeController(Controller):
    def submit(self, amazon: AmazonServices):
        file: UploadedFile = self.request.input("file_registrations")
        s3 = amazon.resource("s3")

        bucket_name = amazon.service_config("s3")["buckets"]["console"]
        exists_path = "/some/bucket/path/to/object"
        s3_obj = s3.Object(bucket_name, exists_path)

        # check if the object exists
        # if so delete it so it can be replaced
        try:
            s3_obj.load()
            s3_obj.delete()
        except ClientError as e:
            if e.response["Error"]["Code"] == "404":
                # The object does not exist.
                pass
            else:
                # Something else has gone wrong.
                raise e

        # put the file in s3         
        try:
            s3_obj.put(Body=file.content)
        except ClientError as e:
            raise e

```
or alternativley esewhere in your app
```python
amazon = self.application.make("amazon")
s3 = amazon.client("s3")
```

I am Interested in any feedback or thoughts on this provider.

Also once this (hopefully) is merged I will update the docs and cookie-cutter to include this

### DIFF
--- a/src/masonite/services/AmazonServices.py
+++ b/src/masonite/services/AmazonServices.py
@@ -1,0 +1,113 @@
+import warnings
+from typing import Any
+from botocore.client import BaseClient
+from ..exceptions import InvalidConfigurationSetup
+
+
+class AmazonServices:
+    """Wrapper for AWS services"""
+
+    def __init__(self, application, config=None):
+        self.application = application
+        self.config = config or {}
+        self._resources = {}
+        self._clients = {}
+
+    def services(self):
+        """list configured service names"""
+        return self.config.keys()
+
+    def client(self, name=None) -> "BaseClient|None":
+        """get a named client interface"""
+        if not name:
+            return None
+
+        try:
+            return self._clients[name]
+        except KeyError:
+            # client not yet initialised
+            client = self._setup_client(name)
+            self._clients[name] = client = client
+            return client
+
+    def resource(self, name=None) -> "Any|None":
+        """get a named resource interface"""
+        if not name:
+            return None
+
+        try:
+            warnings.warn(
+                """The 'resource' interface is no longer being updated by Amazon in boto3.\n 
+                It is recommended to use the 'client' interface instead""",
+                DeprecationWarning,
+            )
+            return self._resources[name]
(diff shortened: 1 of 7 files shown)


josephmancuso:
is this ready and tested?

josephmancuso:
also what service is rgus adding to? file uploading?

josephmancuso:
if this is just proving a proxy to the boto3 library i dont think we should be adding wrapper code to boto3 like this in masonite core code

Issue #752, opened 2023-04-15
use python craft serve error


Mengtuoshisan:
### Describe the bug

(venv) [ayu@fedora myapp]$ python craft serve
Traceback (most recent call last):
  File "/home/ayu/项目/myapp/craft", line 8, in <module>
    from wsgi import application
  File "/home/ayu/项目/myapp/wsgi.py", line 1, in <module>
    from masonite.foundation import Application, Kernel
  File "/home/ayu/项目/myapp/venv/lib64/python3.11/site-packages/masonite/foundation/__init__.py", line 2, in <module>
    from .Kernel import Kernel
  File "/home/ayu/项目/myapp/venv/lib64/python3.11/site-packages/masonite/foundation/Kernel.py", line 35, in <module>
    from ..middleware import MiddlewareCapsule
  File "/home/ayu/项目/myapp/venv/lib64/python3.11/site-packages/masonite/middleware/__init__.py", line 13, in <module>
    from .route.ThrottleRequestsMiddleware import ThrottleRequestsMiddleware
  File "/home/ayu/项目/myapp/venv/lib64/python3.11/site-packages/masonite/middleware/route/ThrottleRequestsMiddleware.py", line 4, in <module>
    from attr import has
ModuleNotFoundError: No module named 'attr'

### OS

Linux

### OS version

Fedora 38

### Browser

edge

### Masonite Version

4.17.4

johnwilson:
@Mengtuoshisan  You can fix this by running `pip install attrs` in your virtual environment. Not sure if this is a python version issue or if `attrs` needs to be added to the `requirements.txt` file.

ghost:
> @Mengtuoshisan You can fix this by running `pip install attrs` in your virtual environment. Not sure if this is a python version issue or if `attrs` needs to be added to the `requirements.txt` file.

I had the same issue as @Mengtuoshisan here, and installing `attrs` in the venv solved the issue, thought it should probably be added by the base project install?

c3hb:
Ran into the same issue with a fresh install, I commented out the `attr import` to get it running for now as it doesn't look like it was being used in the module.

eaguad1337:
same here, I solved it with `pip install --upgrade attrs`, maybe it should be on requirements.txt

josephmancuso:
https://github.com/MasoniteFramework/masonite/commit/84e10358b91a71a61e97df3adc78bb057aba7c44

PR #763, opened 2023-09-25
fix: add missing dependency 

eaguad1337:
New installations fails because `attrs` is not added to setup.py. 

Reference: #752


josephmancuso:
i'm going to actually fix this with this commit https://github.com/MasoniteFramework/masonite/commit/84e10358b91a71a61e97df3adc78bb057aba7c44

PR #779, opened 2023-12-14
fix: delete sqlite db

eaguad1337:
ignore sqlite db and delete from repo


josephmancuso:
why is this? it will fail all the tests right

josephmancuso (changes requested):
this will fail tests removing this from repo right?

PR #787, opened 2024-01-08
fix: force fallback route to be evaluated at the end

### LINKED ISSUE #786: Fallback route should be evaluated at the end

Routes are evaluated in order and if you have a `Route.fallback()` before more routes, these routes will never be evaluated.

_No response_

Make Route.fallback to be evaluated at the end. 

_No response_

- [ ] Yes

### Anything else ?

_No response_

eaguad1337:
Fix #786

### DIFF
--- a/src/masonite/routes/Router.py
+++ b/src/masonite/routes/Router.py
@@ -14,6 +14,9 @@ class Router:
         from .HTTPRoute import HTTPRoute
 
         for route in self.routes:
+            if route.is_fallback:
+                continue
+
             if route.match(path, request_method, subdomain=subdomain):
                 return route
 
@@ -33,6 +36,11 @@ class Router:
         if not matched_methods:
             return None
 
+        # check if there is a fallback route
+        fallback_route = self.get_fallback_route()
+        if fallback_route:
+            return fallback_route
+
         # if alternative methods have been found, check if current request method is OPTIONS
         # to build a proper response else build a method not allowed response
         if request_method == "OPTIONS":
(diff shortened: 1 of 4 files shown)


src/masonite/routes/Router.py
```diff
@@ -49,6 +57,24 @@ def preflight_response(app):
...
+        This method is used to get the fallback route from the list of routes.
+
+        The fallback route is the one where the `is_fallback` attribute is True.
+        If there are multiple fallback routes, it returns the first one.
+        If there are no fallback routes, it returns an empty list.
+
+        Returns:
+            fallback_route (Route): The fallback route if it exists, otherwise an empty list.
+        """
+        fallback_route = list(filter(lambda r: r.is_fallback, self.routes))
```
josephmancuso, line 71: i would imagine this would be a pretty slow way to get a single route. filtering a list just to find the fallback route ..

can we just set this on the router class and just do like 

```python
fallback_route = self.fallback_route
```

tests/integrations/web.py
```diff
@@ -21,9 +21,8 @@
     Route.view("/test_view", "test_view", {"show": "111"}),
     Route.get("/api/uploads/", "WelcomeController@test").middleware("throttle:api"),
     Route.any("/any", "WelcomeController@any"),
+    # Route.fallback("WelcomeController@fallback")
```
josephmancuso, line 24: why is this commented out?

tests/integrations/web.py
```diff
@@ -21,9 +21,8 @@
...
 ]
 
 Broadcast.routes()
-Auth.routes()
-
-Route.fallback("WelcomeController@fallback")
```
josephmancuso, line 29: why is this removed?

josephmancuso (changes requested):

src/masonite/routes/Router.py
josephmancuso: actually nvm its only if theres no route found so its probable fine

PR #792, opened 2024-03-07
fix: add load module error message

eaguad1337:


josephmancuso (approved):

josephmancuso:
i think this is fine. i dont remember entirely why we didnt always raise an exception but we'll figure it out i guess lol

PR #798, opened 2024-04-28
Fixed errors always empty

circulon:
Not really sure how this has stayed unnoticed for so long.

This fixes the "errors" put into session flash being removed at the beginning of the session and put into a view variable called "bag" for some reason.

The upshot of this is that the examples in the Validation documentation never work and the `ShareErrorsInSessionMiddleware` is always empty as it is loaded AFTER the session is established. 

Not really sure how to add a test for this.


circulon:
@josephmancuso  @eaguad1337 
has there been any evaluation of this fix?

This issue may be affecting many devs without them even realising it

josephmancuso:
i dont think we can do this because the documentation says to use the bag helper .. and if we modify this to be done right i think it'll be a breaking change

circulon:
Hi @josephmancuso yeah you are right this would revert to the old behaviour without a MessageBag.
That said the current behaviour is still broken.

I have worked out the issue and am in the process of addressing it.
I will alsp add tests for the Session middleware.

here is what I have so far:
> https://docs.masoniteproject.com/features/validation#displaying-errors-in-views
>Shows to use the  `session().get('errors')` method which always returns `None` 
>The `errors` variable in the View is only referenced as part of the `SharedErrorsInSession`

Additionally:
 It would be nice to have the `errors()` helper to be backwards compatible and return a MessageBag instance

Current errors in view behaviour:
```python
# file: index.html
...
session errors: {{ session().get('errors') }} {# always returns None #}                  
<br>                
session bag: {{ session().get('bag') }} {# always returns None #}
{# just proving there is no session var called 'bag' #}                  
<br>                  
view helper errors() : {# {{ errors().any() }} #} {# throws "errors() is undefined (when uncommented) #}                 
<br>                  
view helper bag() : {{ bag().errors() }} {# contains messages expected in errors() var #}
 ...
```

Expected behaviour:
```python
# file: index.html
...
session errors: {{ session().get('errors') }} {# expected to return None, an empty dict or even a MessageBag instance #}                  
<br>                
session bag: {{ session().get('bag') }} {# always returns None #}
{# just proving there is no session var called 'bag' #}                  
<br>                  
view helper errors() : {{ errors().any() }}  {# Expected value: True #}                 
<br>                  
view helper bag() : {# {{ bag().errors() }}  #} {# Throws 'bag()' is undefined  (qhwn uncommented) #}
 ...
```

I hope this clarifies things

PR #811, opened 2024-07-25
Fix/redis cache issues

circulon:
Refactored Redis Cache Driver

- Fixed internal cache not always loaded on first cache access
- added ability to define “timeout” for for key expiry
- - defaults to a ttl of 1 month.  
- - Previous hardcoded value was 10 years!
- correctly store and unpack int types
- fix internal cache not removed if store was flushed
- fixed an issue where an imternal cache key would not be updated if it existed, but would be updated in the Redis store
- added expiring of internal cache keys

### DIFF
--- a/src/masonite/cache/drivers/RedisDriver.py
+++ b/src/masonite/cache/drivers/RedisDriver.py
@@ -6,7 +6,6 @@ if TYPE_CHECKING:
     from redis import Redis
 
 
-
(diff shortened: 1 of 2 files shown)


circulon:
@goophps
As you are using Redis would you have a look at this for a second opinion please?

josephmancuso (approved):

Issue #816, opened 2024-08-22
Session errors and View 'errors()' helper always empty


circulon:
### Describe the bug

The docs say that you can get any errors in 2 different ways.
```python
{{ session().get("errors") }}
# or
{{ errors().all() }}
```
Neither of these currently work as the SessionMiddleware does the following
> gets the session 'errors' (if any) which exist in the session flash and so are erased after the 'get'
> adds a helper to the View called 'bag()'. which is not documented or intuitive
Additionally:
This impacts the `SharedErrorsInSessionMiddleware' because the session flash 'errors' key has already been removed by the SessionMiddleware

Current behaviour:
snippet from template: index.html
this assumes you have added errors in the conrroller and expect to display with them
in the View template

```python
...
session errors: {{ session().get('errors') }} {# always returns None #}                  
<br>                
session bag: {{ session().get('bag') }} {# always returns None #}
{# just proving there is no session var called 'bag' #}                  
<br>                  
view helper errors(): {# {{ errors().all() }} #} {# throws "errors() is undefined when uncommented #}                 
<br>                  
view helper bag(): {{ bag().all() }} {# prints the messages expected to be in errors() helper #}
 ...
  ```

### Expected behaviour

As per the docs 
https://docs.masoniteproject.com/features/validation#displaying-errors-in-views
I would expect the following

```python
...
session errors: {{ session().get('errors') }} {# shows  whatever messages I put in there #}                  
<br>                
session bag: {{ session().get('bag') }} {# always returns None #}
{# just proving there is no session var called 'bag' #}                  
<br>                  
view helper errors(): {{ errors().all() }}  {# prints all the messages in the message bag #}                 
<br>                  
view helper bag(): {# {{ bag().all() }} #} {# throws "bag()" is undefined when uncommented #}
 ...

Additionally: 
if `SharedErrorsInSessionMiddleware` is used
```python
# in template index.html
...
shared session errors: {{ errors.all() }} {# should show the same messages as the errors() view helper #}
...
```

### Steps to reproduce the bug

as noted issue description above for a View template

### OS

macOS

### OS version

Monterey 12.6

### Browser

Chrome 127.0.6533.120

### Masonite Version

4.20.0 with additional fixes from 4.0 branch

### Anything else ?

Just a thought here but as we have this `errors()` helper in the View, 
isnt the SharedErrorsInSessionMiddleware providing a var in the View calles `errors`
kind of redundant as ther provide the same content in a MessageBag?

PR #817, opened 2024-08-22
Fix session errors and View errors() helper always empty

circulon:
@josephmancuso 

Sorry I broke this PR badly and had to recreate it so we have lost the chat history ;(. 

This PR addresses #816: 
> moves the MessageBag helper in the View from 'bag()'
back to its documented expected value 'errors()'
> adds the "errors" key back into the session for backwards compatibility.

One thing I would ask is:
> Is there any reason not to have the errors added back to the session as a MessageBag?
> This makes SharedErrorsInSessionMiddleware a bit redundant... 

Happy to discuss

will add Tests for SessionMiddleware soon

### DIFF
--- a/src/masonite/middleware/route/SessionMiddleware.py
+++ b/src/masonite/middleware/route/SessionMiddleware.py
@@ -16,13 +16,15 @@ class SessionMiddleware(Middleware):
         request.app.make("response").with_success = self.with_success
         request.app.make("request").session = Session
 
-        # TODO: Remove in Masonite 5
-        errors = Session.get("errors") or {}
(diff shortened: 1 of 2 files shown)


circulon:
@josephmancuso 
I think this is ready now ;)

Looks like this issue has been hiding since v4.16.4 around commit 594d50b0

circulon:
@josephmancuso  @eaguad1337 
could you review this so we can get a new release out with all the current fixes (including this one)

Cheers

circulon:
@josephmancuso  LGTM please and tagged

This is impacting multiple areas of development.

Cheers

josephmancuso:
this is actually a pretty smart solution

